PR #180, opened 2026-08-18
feat(grading): populate summary.wow analytics

hyeonsangjeon:
## What

Three of the four `summary.wow` analytics fields now carry data. `src/types/grade.ts` has declared `by_sector`, `by_rubric_category`, `score_density_histogram` and `rubric_severity_curve` since the WOW dashboard landed, and `SectorHeatmap.tsx`, `ScoreDensityHistogram.tsx` and `RubricSeverityCurve.tsx` have rendered their empty states ever since — `_compute_summary` in `step8_grade.py` never emitted them.

| field | now |
|---|---|
| `by_sector` | `task_count`, `avg_pct`, and the three pass rates, per sector |
| `score_density_histogram` | all ten deciles, empty buckets included |
| `rubric_severity_curve` | items grouped by rubric weight, sign-aware |
| `by_rubric_category` | stays `{}` — see below |

### Design points worth reviewing

- **Sector rates roll up.** The run-wide summary and every sector bag are folded by one shared `_tally_item`. A second copy of those rules would drift the first time someone edited one, and the dashboard would then show sector rates that don't roll up to the header above them.
- **Counts add up.** `by_sector` is scoped to graded tasks, so `sum(task_count) == graded_tasks`; a blank sector buckets to `Unknown` rather than disappearing from a total that's expected to add up. Same for the histogram.
- **The histogram labels are a contract.** `bucketFromPct` in `ScoreDensityHistogram.tsx` buckets `pct` client-side when a grade predates the field, so a run that emits it and one that doesn't must land in the same bars. `100.0` closes into the last bucket.
- **The severity curve counts `model_did_right`, not `verdict == 'pass'`.** GDPVal rubrics carry negative-weight anti-criteria where a `pass` verdict means the model *did* the prohibited thing, and this curve deliberately spans both signs — a raw verdict count would invert exactly the points the chart exists to show.
- **The curve is omitted for pre-sign-aware payloads.** Grades written before `core/grader.py` emitted `model_did_right` would plot a flat zero line across every weight: a chart asserting a total failure that never happened. A single rate can absorb that gap; a curve can't, because its shape is the claim. 4 of the 16 checked-in payloads are in this category.
- **`by_rubric_category` is deliberately empty.** The rubrics carry no category taxonomy — an item has an id, a criterion string and a weight, nothing that groups items. Populating it would mean inventing a taxonomy and presenting it as measurement. `SectorHeatmap.tsx` already falls back to the per-sector rates.
- `judge_error_rate` keeps `canonical_rate` (validated by `grade_payload.py`, cross-checked by `step9_merge_shards.py`); only the new fields use the plain rounding helper.
- `step9_merge_shards.py` recomputes the merged summary rather than combining shard summaries, so a sharded 220-task run gains these fields with no merge-side change.

## Also: a red test on `main`

`test_grade_workflow_rc7_requires_valid_committed_partial` asserted the `approve-paid` job's `if` as an exact literal. #179 added the `approval_inherited != 'true'` conjunct and a YAML block scalar, so **this test has been failing on `main` since #179 merged.**

Nothing caught it: no workflow runs the `batch-runner` pytest suite on pull requests. `agentic-sandbox-preflight.yml` is the only workflow that invokes `pytest` at all, and it exercises the sandbox preflight. Flagging that gap rather than fixing it here — a CI job is a separate change with its own review surface.

The assertion now compares whitespace-normalized expressions (strict about the condition, indifferent to YAML reflow) and pins the `grade` job's condition exactly, replacing four substring probes that a weakened gate would still satisfy. Confirmed load-bearing by mutation: dropping the `approval_inherited` conjunct from the skip-escape — which would let a paid run start with neither a click nor a proof — fails the test. `grade-run.yml` is correct and unchanged; only the test moved.

## Verification

- **Behavior-identical on pre-existing fields.** Ran `_compute_summary` from `origin/main` and from this branch over the same 16 checked-in grade payloads. Every previously-emitted field matched on all 16 (`pre-existing field drift : 0`).
- **New fields self-consistent** on all 16: `sum(by_sector[*].task_count) == graded_tasks == sum(histogram[*].count)`.
- `mypy core/ step8_grade.py --ignore-missing-imports` — identical error set before and after.
- `ruff check` — identical findings before and after.
- 464 tests pass (`test_step8_grade`, `test_grade_schema`, `test_step9_merge_shards`, `test_narrative_grade_integration`, `test_grader`), including 18 new ones.

No grade payload was rewritten, no data file touched, and no workflow dispatched.

## ⚠️ Merge timing

`step8_grade.py` is inside the `compute_grader_source_hash` source set, so **merging this changes `grader_source_hash`** and therefore the shard partial path `data/grades/_shards/<stem>/`.

A chunk relay that is mid-flight when this merges will not find its own previous partial, will fail the `approval_inherited` check and fall back to a fresh Environment approval, and its shard set will fail the cross-shard `grader_source_hash` invariant at `step9_merge_shards.py:122`.

**This must merge between runs, not during one.** Run `32166049814` is in progress as of this writing.


hyeonsangjeon:
Brought onto current `main` (`ab446c8`) before merging. Recording what changed since review, since the base moved 30 commits.

**Not a history rewrite.** This repo prohibits force-push, so the branch was synced by merging `main` into it rather than rebasing. The PR squash-merges, so the commit that lands on `main` is identical either way.

**Two conflicts, both in `CHANGELOG.md`:**

- `### Added` — additive. The wow-analytics entry sits above `main`'s four.
- `### Fixed` — **the same fix, written twice.** `main` and this branch made byte-identical changes to `test_grade_workflow_rc7_requires_valid_committed_partial`, so git merged the code cleanly and only the prose collided. Kept `main`'s entry; dropped this branch's, which additionally stated that PR-time backend test coverage was absent. `backend-tests.yml` has landed on `main` since, so that sentence is no longer true.

The operational note now records that the merge did happen between runs, rather than warning that it must.

**Re-verified after the merge, not just before it.** The behaviour evidence in the description was gathered against a `main` that has since moved, so it was re-run:

- `main`'s 30 new commits touch **no production grading code** — two test files only (`test_step8_grade.py`, `test_silent_corruption_fixes.py`). The shared code path is unchanged, so the original comparison still holds.
- Re-ran `_compute_summary` from `main` and from the merged tree over all 18 checked-in grade payloads: **0 changes to any published figure**. All 47 new-field transitions are `empty -> populated`; nothing non-empty moved. `dummy_gpt5_baseline.json` raises the same `KeyError: 'pct'` on both sides — pre-existing, and that payload is already hidden from the dashboard.
- New fields self-consistent on 17/17 real payloads: `sum(by_sector.task_count) == sum(histogram.count) == graded_tasks`, 10 buckets each.
- 451 tests pass locally on Python 3.10.12 (`test_step8_grade`, `test_grade_schema`, `test_step9_merge_shards`, `test_grader`); `pytest` green in CI.
- Merged tree differs from `main` in **exactly three files**, and `step8_grade.py` is byte-identical to this branch's own version — no merge artifact. `data/grades/` and `.github/` are byte-identical to `main`.

**Merge-timing hazard cleared.** `step8_grade.py` is inside `compute_grader_source_hash`, so this changes the hash. Checked immediately before merging: no `grade-run` in flight, no queued or in-progress runs repo-wide. The 220-task relay finished.

PR #197, opened 2026-08-22
feat(grading): report shard relays that stop without finishing

hyeonsangjeon:
## What this closes

`#194` gave a shard that finds a short union a way to stand down politely: exit `75`, let whichever sibling finishes last do the merge. That is right for the normal case and it removed a large class of false red runs. It also widened a hole that was already there — if one shard never finishes, every surviving sibling defers, nobody merges, and no final grade is written.

The gap is narrow and specific:

| shard outcome | reported today? |
|---|---|
| crashes | ✅ its own run goes red |
| finishes, union short | ✅ defers on purpose, sibling merges later |
| **ends green without finishing** | ❌ **nothing anywhere** |

The third row happens when the auto-resume relay's `gh workflow run` dispatch fails — rate limit, expired token, a transient API error — and the run still exits successfully. The relay is broken and every signal is green.

It can't be closed inside the merge step. Shards are independent `workflow_dispatch` runs rather than a matrix, so no job can `needs:` all of them and there is nowhere to hang "everyone is done, now merge". This watches from outside instead, on a three-hour schedule.

## How it decides

**Liveness comes from git, not the workflow API.** A working shard commits its slice at the end of every chunk, so the newest commit touching `data/grades/_shards/<stem>/` is a direct record of progress. The API alternative would have to match dispatch inputs against the display title `Run GDPVal Grade Pipeline`, which every shard of every run shares. Eight hours of silence is the default threshold — roughly double a full chunk, late enough never to interrupt a slow-but-working run.

**Missing ids are recovered by hash, not by name.** Each grading config's pinned `rerun_identity.task_ids` is hashed and matched against the `expected_ordered_task_ids_sha256` the shards themselves recorded, so a match is verified rather than assumed. With no match it degrades to shortfall counts derived from the stride profile — still enough to name the shard to re-dispatch.

**Not-yet-published is reported, never failed on.** The canary-first procedure deliberately runs shard 0 alone for hours. Liveness git can't read (shallow clone, uncommitted tree) is reported as `unknown` rather than guessed from mtimes, which after a CI checkout would make every stem look brand new.

Sample of the failing output:

```
[stalled] <stem>
    5/6 tasks in and nothing has been committed for 514.45h
    shard 001 still owes 1 task(s)
```

## Safety

Strictly read-only — it never dispatches, merges, or writes. Its whole output is a red run and a job summary. The happy path is untouched: the last shard still merges, deferring shards still exit green.

## Tests

26 new tests. The two values the sweep re-derives from `step8_grade` — stride ownership and the ordered-id hash — are checked against the real implementations rather than transcribed constants, so drift fails a test instead of quietly blaming the wrong shard in a report someone acts on. One test asserts the shipped 220-task config is recoverable by the hash path, since that is the run this exists to watch.

Full suite on this branch: `3 failed, 3357 passed, 6 skipped` — the three failures are the known local-environment ones (pinned Azure SDK versions, absent `pdfplumber`) and are identical on `main`. Collection delta is exactly the 26 added.

## Note on the freeze

`.gitignore` needed the new script added to the `batch-runner/scripts/` allow-list. Everything here is outside `compute_grader_source_hash` and unread by `grade-run.yml`, so it is safe to land while shards are in flight.

### DIFF
--- a/.github/workflows/shard-relay-sweep.yml
+++ b/.github/workflows/shard-relay-sweep.yml
@@ -1,0 +1,90 @@
+name: Shard Relay Sweep
+
+# Watches for a sharded grade run that stopped without producing a final grade.
+#
+# `#194` let a shard that finds a short union stand down (exit 75) and leave the
+# merge to whichever sibling finishes last. That is right for the normal case,
+# and it removed a large class of false red runs. It also widened a hole: if one
+# shard never finishes, every sibling defers, nobody merges, and no final grade
+# is written. The shard that *crashes* reports itself -- its own run goes red.
+# The shard that ends green without finishing does not, and that is the case
+# this sweep exists for. It happens when the auto-resume relay's
+# `gh workflow run` dispatch fails (rate limit, expired token, a transient API
+# error) and the run still exits successfully.
+#
+# It cannot be closed inside the merge step, because shards are independent
+# `workflow_dispatch` runs rather than a matrix: no job can `needs:` all of
+# them. So this watches from outside.
+#
+# Strictly read-only. It never dispatches a shard, never merges, never commits.
+# Its whole output is a red run and a job summary.
+
+on:
+  schedule:
+    # Every three hours, off the hour. The sweep tolerates eight hours of
+    # silence before calling a relay stopped, so this cadence finds a broken
+    # relay within a working day without ever racing a healthy chunk.
+    - cron: "37 */3 * * *"
+  workflow_dispatch:
+    inputs:
+      stale_after_hours:
+        description: "Silence tolerated before a relay is called stopped"
+        required: false
+        default: "8"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: shard-relay-sweep
+  cancel-in-progress: false
+
+jobs:
+  sweep:
+    runs-on: ubuntu-24.04
+    timeout-minutes: 15
+    env:
+      PYTHONDONTWRITEBYTECODE: "1"
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        with:
+          persist-credentials: false
+          # Full history, not the default shallow clone. Liveness is read from
+          # `git log -- data/grades/_shards/<stem>`: a working shard commits its
+          # slice at the end of every chunk, so the newest commit touching a stem
+          # is a direct record of progress. A depth-1 clone has no such history
+          # and every stem would report "cannot tell".
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
+        with:
+          python-version: "3.11"
+
+      # The sweep reads JSON and one YAML key. `step9_merge_shards`, which it
+      # borrows the shard loader from, is stdlib-only. Installing the full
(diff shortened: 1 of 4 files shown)


hyeonsangjeon:
CI earned its keep here. The three git-backed tests failed on the runner, and chasing that turned up a defect in the shipped path rather than a test-only one.

`last_commit_at` passed the stem path to git unchanged. The workflow defaults are `--repo-root ..` and `--grades-root ../data/grades`, both relative to `batch-runner/`, so git was asked about a pathspec beginning with `..` from a cwd inside the repo:

```
fatal: '../data/grades/_shards/<stem>' is outside repository at '/…/repo'
```

Exit 128 → `None` → "cannot tell". Every stem would have reported `unknown`, and **the sweep would have been silently useless in the one workflow it exists for — while exiting 0 and looking healthy.**

Running it against the real repository did not catch this and could not: the published run short-circuits at `merged` before liveness is ever read, so the only path I had exercised was the one that never calls git.

Now fixed by making the pathspec relative to the resolved work tree (both ends resolved, so symlinks on either side stop mattering) and running git from that root. Absolute pathspecs are gone entirely, which is also what the runner's git was rejecting. Alongside it: `log.showSignature` neutralised per invocation, and the timestamp taken from the first line that parses rather than the whole stream.

The fixture now asserts its own contract — if git cannot make or read a commit in the temp repo it reports git's return code and stderr, instead of surfacing as a bare `assert None == datetime(...)` somewhere unrelated. And the new test drives the CLI through the relative roots the workflow actually passes; it fails on the previous code.

hyeonsangjeon:
Root cause of the four red runs, and it was not what I kept guessing.

The instrumented assertion reported the failing call verbatim:

```
argv=[git, -c, log.showSignature=false, log, -1, --format=%cI, --, grades/_shards/run-stem]
rc=0  out='2026-08-20T09:00:00Z\n'  err=''
version='git version 2.55.0'
```

git succeeded and returned the right instant. It just spelled it `Z`. git 2.55, which the ubuntu-24.04 runners carry, renders a UTC `%cI` as `...Z`; git 2.34 on my box renders `...+00:00`. Python 3.10 (`backend-tests.yml` pins 3.10.12) only accepts the second, so `fromisoformat` raised, the loop skipped the one correct line, and `last_commit_at` returned `None`.

`None` means "git cannot tell us" — a real state a shallow clone produces, and one the sweep deliberately keeps quiet about rather than guessing from mtimes. So the failure mode in production was not a red run. It was the scheduled workflow exiting 0 every three hours with nothing to report, which is indistinguishable from a healthy repository. Same shape as the pathspec bug in 4fc96a7: the watchdog silently stops watching.

I had dismissed `fromisoformat` early on the assumption that `%cI` always emits `+00:00`. That assumption is why this took four round-trips.

The parsing moved into `parse_commit_stamp` so the regression test feeds it both spellings directly rather than shelling out — a test that runs git can only ever prove the host's side of a version boundary. As a cross-check I re-ran the whole suite behind a stub `git` that appends `Z`: 31 pass on this commit, and `last_commit_at` returns `None` on the previous parser under the same stub.

Also dropped the pathspec probe from the fixture. It never fired, and `_explain_liveness` — which did the actual work here — covers the same ground with the real invocation attached. Kept that one.

Verified: 31 sweep tests, 392 across the shard modules, and the sweep still reports `[merged]` rc=0 against the real repository under both a `Z`-emitting and a `+00:00`-emitting git.

PR #238, opened 2026-08-26
fix(envelope): read the input files instead of half their fingerprints

hyeonsangjeon:
## The claim that was not true

The free gate that stands between this comparison and a bill describes itself, in `core/execution_envelope_preflight.py`, as checking "the fingerprints of every input file against the dataset".

Nothing read a byte of any file.

`check_input_file_versions` compared the first **32** of each fingerprint's **64** characters against the folder name in the file's own path — this benchmark keeps a reference file in a folder named after the start of its fingerprint. The other 32 characters were compared against nothing at all.

## Measured, old code against new, in one run

Every single-character change, at every one of the 64 positions, on all three fingerprints the committed plan pins. Exhaustive rather than sampled, because the failure being fixed was a *range of positions nobody had looked at*.

| | changes tried | went unnoticed |
|---|---|---|
| before | 210 | **70** |
| after | 210 | **0** |

Where the 70 were:

| entry | missed | why |
|---|---|---|
| the dataset fingerprint | 0 | already compared in full, against the catalogue |
| reference file A | 35 | positions 32–63, plus three whole-second-half swaps |
| reference file B | 35 | same |

The three swaps replaced the second half with the tail of the other reference file, of the dataset fingerprint, and of an unrelated file. All three passed.

## What it does now

Finds a copy already on this machine, hashes it, compares all 64 characters. **It never downloads.** Two places are looked at, in this order:

1. a folder named on purpose with the new `--dataset-root` option;
2. the Hugging Face download cache, asked about that one file at that one pinned revision — a question that fetches nothing.

## The part that mattered more than the hashing

What to do when no copy is there. Staying quiet would make "checked and matched" and "never looked" identical in the report — which is the same defect one level up. So each file reports its own state:

| state | meaning | compared |
|---|---|---|
| `read the file` | read off this machine, all 64 agreed | 64 / 64 |
| `folder name only` | no copy; only the folder's half could be compared | 32 / 64 |
| `not checked` | nothing to compare against | 0 / 64 |

The advance check prints this table **whether or not anything is wrong**, and each unchecked file becomes a problem naming the free command that resolves it. This is the rule the repository already used for the Docker probe: `--skip-docker-probe` reports the container "as not measured rather than as ready".

## Two more holes closed

- The folder rule only ran when the folder name was **exactly** 32 characters long. A path of any other shape skipped the comparison entirely and was reported as fine — the one answer that was certainly wrong. It is now `not checked`.
- Where a copy is found decides what a mismatch *proves*. A reference file's path repeats the first half of its own fingerprint, so a copy found there can only be that file, and a mismatch means the written value describes some other file. The dataset's own data file carries no such promise: a folder somebody pointed at may hold a different revision, so that mismatch is reported as exactly that ambiguity instead of as tampering.

## The first version of this was only ever tried on a machine that had the files

It passed here and failed on the build runner, in six tests that have nothing to do with input files. That is worth writing down, because it is the same shape of mistake as the one this branch set out to fix: a condition nobody had run under.

This machine has the pinned revision sitting in its Hugging Face download cache. A fresh runner does not. So the new "no copy of this file is on this machine" notes appeared only there, and six tests asserting *"the thing I am testing is right, and nothing else is wrong either"* tripped over them.

The two findings were sharing one list, and they are not the same finding:

| finding | what it means | same on every machine? | how it clears |
|---|---|---|---|
| **does not match** | the plan pinned the wrong file | yes | fix the plan |
| **could not check** | this machine has not downloaded the benchmark | **no** | one free command |

They now have two names — `InputFileVerification.missing_copies` and `EnvelopePreflight.missing_input_file_problems`. **Nothing was silenced:** both still stop a run, both are still printed together, and `check_input_file_versions` still returns both. The only thing that changed is that a test about something else can name the machine-dependent half and set it aside on purpose.

The set-aside is by **list membership, never by matching text**. A substring filter was the obvious fix and the wrong one: it could also have swallowed a real disagreement, and then a plan pinning the wrong file would have passed all six tests silently. `test_only_files_that_are_absent_are_ever_set_aside` asserts a disagreement can never reach that list.

Two of the new tests run with the download cache emptied — the condition that was never tried.

While here, the cross-reference guard was widened from the top docstring to every docstring in the module, and taught about dataclass fields. It found a stale `` :attr:`problems` `` on its first run.

## Checks

- `pytest`, **same tree, both conditions**:

  | condition | result |
  |---|---|
  | cache present, as this machine is | **4235 passed, 6 skipped** in 578.49s |
  | cache emptied, as a fresh runner is | **4235 passed, 6 skipped** in 542.89s |

  4202 before this branch, plus 33 new tests. The two numbers being equal is the point.
- `mypy core/ --ignore-missing-imports`: **170 errors in 32 files** — the baseline, unchanged
- the three fingerprints in the committed plan were **read off this machine and all matched**; no invented failure
- **no model was called, nothing was downloaded, no approval was granted, no block was lifted.** The gate only refuses more than it did

Specification: §13.13 of `tasks/0822_saturday/TASK_GPT_EXECUTION_ENVELOPE_BENCHMARK.md`.


hyeonsangjeon:
### Why there is no check on this branch

GitHub Actions is in a critical outage (incident opened `2026-08-26T15:11:58Z`). This
branch's run started at `15:25:02Z`, thirteen minutes into it, and ended in
`startup_failure` — "The job was not acquired by Runner of type hosted even after
multiple attempts". No job was ever created, so no test ran; `gh run view --log-failed`
returns nothing. Re-running sat queued for 56 minutes, and a force-push produced no run
at all. The repository's runners were healthy until 14:38, so this is not the branch.

The six real failures on the earlier run (`32981554414`, `c82cf02`) were a genuine defect
and are what this commit fixes: a build server has no copy of the benchmark, so
"could not check this fingerprint" was being filed as "this fingerprint is wrong". They
are now two lists. Both still stop a run.

**This branch is not merged and will not be merged until a run actually executes.**

### What was verified locally instead, in both conditions

The condition CI runs under is a machine with no download cache. That is reproduced here
by emptying `HF_HOME` and by monkeypatching `try_to_load_from_cache` to return `None`.

| Suite | Cache present | Cache emptied |
| --- | --- | --- |
| Full backend suite | 4235 passed, 6 skipped, 45 deselected (578s) | 4235 passed, 6 skipped, 45 deselected (543s) |
| `test_envelope_preflight_reads_the_input_files.py` | 32 passed | 32 passed |
| `mypy core/` | 170 errors in 32 files | baseline, unchanged |

The two totals being equal is the point. That equality is the property the six CI failures
were complaining about, and it is now covered by tests rather than by the machine.

PR #256, opened 2026-08-28
feat(cost): record and show per-task problem-solving and grading cost

hyeonsangjeon:
## What this adds

Every task now carries two receipts, kept apart because they answer different questions: **문제 풀이 비용** (what it cost the model to produce the deliverable) and **채점 비용** (what it cost to grade it). Both flow from the producer's ledger through step 3 and step 6 into `self_report.json` and the published parquet, and out to the dashboard.

### Producer contract, as shipped

- `components[].name` is one of `preprocessing`, `generation`, `self_qa`, `grading`, `perception`, `retry`. Work that had to be redone is filed under `retry`, with the originating stage kept beside it, so two stages that each retried never read as one line.
- `runtime` is **not** a component. It is a sandbox fee, not a model call, so it gets its own line and is never restated as a component — otherwise the lines would double-count against the total.
- Grading receipts are read only from **grade schema 1.4**. Schemas 1.0–1.3 are still read for everything else and simply carry no cost.

### Four states, kept distinct

`complete` / `partial` / `unpriced` / `not_run` are never collapsed. An unpriced receipt does not become a zero, and a stage that never ran does not become a missing one.

## Fixes found while reconciling against the merged producer

Three defects on the publication path, all of which would have surfaced only after a paid run:

1. **Dead digest check.** Steps 3 and 6 looked for the ledger under `workspace/upload/`, but step 2 writes it to `workspace/cost_ledger_<condition>.jsonl`. `verify_cost_ledger` accepts a missing file, so the check silently never ran.
2. **Nothing staged the ledger.** No step copied the export into the upload directory, and `self_report.json` would have declared the producer's per-condition name while publication pins `cost_ledger.jsonl` — the upload would have failed its own validation either way. The published name is now pinned once at the projection layer and read from there by both `hf_publication` and the staging step.
3. **Stale allowlist mirror.** `step7_upload_hf.sh` restates all three pattern lists and aborts if they differ from the module's. Adding `cost_ledger.jsonl` to the module without the mirror would have stopped *every* upload, including experiments unrelated to cost. Mirror updated, and now covered by tests on both sides.

Two more found by reading the shipped spec (`tasks/0828_friday/TASK_PER_TASK_COST_RECEIPTS.md`) against this side:

4. **Invented reason slugs.** The fixtures used `price_table_missing` and `runtime_price_missing`. `missing_reasons` is a closed enum in `grade.schema.json`, and neither value is in it — so the tests that appeared to cover the unpriced path were asserting on strings the pipeline cannot emit. Replaced with the real values they meant.
5. **Untranslated reasons on screen.** The eight real values reached the reader as-is: `미가격 사유: price_missing, usage_absent`. The one line whose whole job is to say why there is no number was the one line that could not be read. Each now has a Korean label, and a test reads the enum out of the schema and the label keys out of `src/lib/cost.ts` and requires them to match — a ninth reason fails at `node --test` rather than appearing as a slug. An unknown reason still prints raw, on purpose: dropping it would lose the only word saying what went wrong.

Also: the runtime line no longer appears on tasks that never opened a sandbox. The producer sums runtime into a `Decimal` starting at zero, so `runtime_cost_usd` is present on every receipt; gating the line on presence put `실행 환경 $0.0000` on every task.

## Verification

All free — no paid model calls, no real grading, no smoke or pilot run.

| Check | Result |
|---|---|
| `pytest` | 5575 passed, 6 skipped, 0 failed |
| `npx tsc --noEmit` | clean |
| `npm run test:aggregate` | 176 passed, 0 failed, 1 skipped¹ |
| `npm run build` | ✓ |
| Playwright suites (6) | all pass, including a new `cost-receipts` suite |

¹ Pre-existing environmental skip: Ruby is not installed locally.

The browser fixtures now enter in the shape `core/cost_receipts.py` actually writes and pass through the real projector, so they exercise the placeholder-to-absence conversion instead of assuming it.

## Note on history

This branch was rebased onto current `main`. Rather than force-push over the previously pushed tip, its four pre-rebase commits — the same work on an older base — are reconciled with an `ours` merge. The tree is byte-identical to the rebased one, the push was a fast-forward, and nothing on the remote was discarded.


hyeonsangjeon:
세션 A의 PR #257이 `main`에 병합됐다 — 병합 커밋 **`db96c92`**. 이제 이 PR을 최신 `main` 위로 rebase할 수 있다.

**파일은 한 개도 안 겹친다** (#257의 21개 vs 이 PR의 23개, 교집합 공집합). 텍스트 충돌은 안 날 것이다.

다만 **통합 전에 확인이 필요한 의미 불일치가 하나 있다.** 세션 B 파일은 건드리지 않았고, 프로젝트 카드에 호환성 요청으로 적어 뒀다.

`scripts/cost-receipt.mjs:188-191`이 `components[].name`을 중복 불가 키로 다루는데, 생산자는 `name`을 **라벨**로 쓴다. 재시도가 두 번 이상이면 `retry`가 두 줄 이상 나온다. 실제 실행 결과다.

```
emitted names: ['generation', 'retry', 'retry']   → DUPLICATES: True

  name=generation   stage=generation   retry_kind=none
  name=retry        stage=generation   retry_kind=semantic
  name=retry        stage=self_qa      retry_kind=infrastructure
```

이대로 만나면 줄 하나가 아니라 **그 작업의 영수증 전체가 `fail()` 된다.** Self-QA 재시도가 있는 실행에선 흔하게 발생한다.

실제 식별자는 `(stage, retry_kind)` 조합이고 둘 다 같은 줄에 있으니, 유일성 검사를 걷어내고 `name`으로 묶어 합산하는 쪽을 제안한다. 병합된 JSON Schema에도 `components` 유일성 제약은 없다 — 중복 허용이 생산자 쪽 의도된 계약이다.

판단이 다르면 카드에 적어 달라. 임의로 정하지 않았다. 자세한 근거와 나머지 인계 사항은 프로젝트 카드 "세션 B 인계" 절에 있다.

PR #257, opened 2026-08-28
feat(cost): record what each task actually cost to solve and to grade

hyeonsangjeon:
Records what each task actually cost — once to solve, once to grade — as a
receipt written at the moment the money was spent, rather than a number
reconstructed afterwards from whatever survived.

This is Session A of a two-session split. Session B (#256) consumes what this
produces; the contract between them is frozen on the Project #5 card. Session A
merges first.

---

## ⚠️ Do not merge while a grading run is in flight

At the time of writing, `exp_gold_baseline` is mid-flight: shards 0 and 2 of 3
have committed to `main`, and shard 1 (run `33187134567`) is still running.
**Merging this PR before that run's merge step completes will destroy it, after
the money has already been spent.** Two independent reasons:

1. **Schema version.** Step9 pins each shard's `schema_version` against the
   current contract constant, not just against its siblings. This PR takes that
   constant from `1.3` to `1.4`. Verified directly against the two real shard
   files now on `main`:

   ```
   ShardMergeError: shard merge identity mismatch for schema_version:
   s0='1.3', current contract='1.4'
   ```

2. **`grader_source_hash`.** `compute_grader_source_hash` digests
   `step8_grade.py`, **every** `core/**/*.py`, and `schemas/grade.schema.json`.
   This PR adds two core modules and edits four more, plus step8 and the schema.
   Shard 1 running against a post-merge checkout would record a different hash
   from its two already-committed siblings, and step9 would refuse the merge —
   again only after the spend.

Neither is a defect introduced here; both are the existing guards doing their
job. They just make merge timing load-bearing. **Wait for `data/grades/` to hold
the merged `exp_gold_baseline` final, then merge.**

## What gets recorded

A receipt is written per `(task, bucket)`, where bucket is `problem_solving` or
`grading`. Every model call is **reserved before it goes out** and **settled when
the response comes back**, so a call that was billed but whose answer was thrown
away still appears — and a call that failed before it ever left the process does
not.

The ledger is append-only SQLite with `synchronous=FULL`, keyed by a
deterministic `call_id` = `sha256(run_id|task_id|stage|retry_kind|attempt_index|
sequence)`. That primary key is the whole concurrency story: two processes
racing to settle the same call cannot double-bill it, and a resumed round
re-walking its own history re-reserves rows it already has instead of resetting
them.

Stages are `preprocessing`, `generation`, `self_qa`, `grading`, `perception`.
Retry kinds are `none`, `semantic`, `infrastructure`, `resume`,
`internal_recovery`.

## Refusing to invent a number

`estimated_cost_usd` is a number **only** when status is `complete` — every call
settled, every model priced. Otherwise it is `null`, and whatever *is* known
goes to `known_cost_usd` as a floor. The distinction that matters: a task with
an unpriced model is not a task that cost `$0`.

Specifically:

- **No near-miss pricing.** A price is matched on the exact
  `(provider, resolved_model)` the *response* reported, never the deployment
  name that was requested. No fallback to a similar model — `price_missing`,
  `partial`.
- **No double-billing of cached input.** Billable input is
  `input_tokens − cached_input_tokens`.
- **No double-billing of reasoning.** `reasoning_billed_as` is
  `output | separate | unknown`; `unknown` refuses to guess.
- **A request that never came back holds the receipt open.** A timeout may well
  have been billed, so the reservation stays unsettled and the receipt is
  `partial` with `call_reachability_unknown` — not written off as free.
- **Shared runtime is never allocated.** A container billed per-hour across many
  tasks is not divided by task count to produce a plausible-looking number. The
  task receipt stays `partial`. Only directly attributable runtime is priced,
  and it is reported as `runtime_cost_usd`, separately from model cost.

## The paths that are not metered, and what they say instead

Meter coverage is not universal, and the uncovered paths say so rather than
reporting zero:

| Path | Receipt | Reason |
|---|---|---|
| Hardened / agentic substrate | `unavailable` | `stage_unsupported`, plus an explicit console line |
| `agentic_sandbox_v2` | `not_run` | Its executor raises on model clients, credentials, model names, prompts and custom backend factories — it is structurally incapable of spending money |
| Native Codex | — | No execution mode exists |

**Known gap, stated rather than papered over:** there is no regression test on
the `hardened_requested` branch itself. No hardened end-to-end harness exists in
the suite (`hardened=True` is a defined parameter that is never exercised), and
building one needs the signed approval gate. The branch is three lines and is
readable at `step2_run_inference.py:3251-3302`, but it is not pinned by a test,
and a future edit could silently start metering — or silently stop.

## Contract answer for Session B: `components[].name`

Session B asked on the card for the exact slug vocabulary, having assumed
`{generation, self_qa, retry, runtime, grading, perception}` with aliases and an
unknown-slug fallback. The producer previously emitted **no `name` key at all**.
Resolved additively — nothing renamed, nothing removed, no Session B file
touched:

```
COMPONENT_NAMES = (preprocessing, generation, self_qa, grading, perception, retry)
```

`name` is derived: first work carries its stage's name, and anything that had to
be done again is `retry`, because "what did retrying cost" is the question a
reader asks. Neither underlying field is lost — `stage` and `retry_kind` still
travel on the same line.

Five of Session B's six match exactly. `preprocessing` is extra, which their
fallback already handles. **`runtime` is deliberately not emitted**: runtime fees
are not model calls, they are reported as `runtime_cost_usd`, and a component
line carrying them would be counted twice by any reader that sums the lines and
then adds the runtime total. A test pins that absence.

## Published rows, not the live ledger

Step8's `summary.grading_cost` is summed from the task rows via
`CostReceipt.from_dict` — not from the ledger it just wrote. That one choice is
what makes shard merging correct for free: step9 recomputes the merged summary
from merged rows, and the arithmetic agrees by construction. Step9's
cross-shard check compares integer `model_calls`, never rounded decimal money,
so it catches a shard whose receipts silently failed to arrive without
manufacturing a float-comparison flake.

## Nothing sensitive reaches the ledger

No prompt text, no response bodies, no keys, no reasoning traces. Request
identifiers are accepted **only** as SHA-256 digests — passing anything that
isn't one raises. The export is byte-reproducible, so the digest published
beside a grade can actually be checked, and `verify_export` does check it.

## Tests

`5171 passed, 6 skipped, 45 deselected` on this head — full suite, exit 0. 512
of those cover the cost surface directly, across five new files:

- `test_cost_receipts_keep_every_call_that_happened.py` — a call that happened stays recorded
- `test_cost_receipts_never_invent_a_number.py` — the refusals above
- `test_cost_receipts_reach_the_pipeline_boundaries.py` — real step2/step8 wiring
- `test_cost_receipts_survive_resume_and_shard_merge.py` — preservation across rounds and merges
- `test_cost_metering_records_what_the_client_sent.py` — the client wrapper

All 16 scenarios required by `tasks/0828_friday/TASK_PER_TASK_COST_RECEIPTS.md`
§8.2 have a test. `test_azure_ai_step2_wiring.py` now looks *through* the
metering wrapper, which pins every executor client slot as metered — a future
unwrapped client fails a test instead of silently going unbilled.

**Every test runs on mocked responses.** No paid model call, no Azure login, no
grading or regrade, no Smoke/Pilot/Full run was performed in this work.

## Also in this PR

`tasks/0828_friday/TASK_PER_TASK_COST_RECEIPTS.md` — the written spec, which the
card referenced but which existed in no checkout. It is written to be
independent of any particular model or provider.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,6 +80,9 @@ tasks/0822_saturday/*
 !tasks/0822_saturday/TASK_PIN_AZURE_RESOURCE_IDENTITY.md
 !tasks/0822_saturday/TASK_AGENTIC_SANDBOX_V2_FOUNDATION.md
 !tasks/0822_saturday/TASK_NATIVE_CODEX_RUN_PATH.md
+!tasks/0828_friday/
+tasks/0828_friday/*
+!tasks/0828_friday/TASK_PER_TASK_COST_RECEIPTS.md
 # The grading-rebuild specs and reports here predate the `tasks/**` rule, so
 # they stay tracked without an entry. A *new* file among them does not -- git
 # will not un-ignore a path inside an ignored directory, which is why the
--- a/batch-runner/core/cost_metering.py
+++ b/batch-runner/core/cost_metering.py
@@ -1,0 +1,539 @@
+"""Turning a provider client into one that keeps its own receipts.
+
+The pipeline calls models from a dozen places — a subprocess runner, a
+container, a self-check, four judges, two perception readers. Threading a
+ledger through every one of those signatures would touch most of the codebase
+and would still miss whichever path was added last.
+
+So the ledger is attached to the *client* instead. :meth:`CostRecorder.meter`
+wraps whatever the provider factory returned, and from then on every
+``chat.completions.create`` and ``responses.create`` that goes through it
+reserves a row before the request and settles it after the reply, without the
+calling code knowing. What the calling code does supply is *whose* call it is —
+:meth:`CostRecorder.attributed` marks a block as belonging to one task at one
+stage, and calls made inside it are filed there.
+
+Metering is opt-in, one client at a time. A path nobody wrapped records
+nothing, and its receipt says ``stage_unsupported`` rather than quietly
+reporting a smaller bill. That is the intended behaviour for execution paths
+that have no confirmed model call to meter: silence is reported as silence.
+
+Nothing here contacts a provider. It sits between code that does and the
+ledger.
+"""
+
+from __future__ import annotations
+
+from contextlib import contextmanager
+from contextvars import ContextVar
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Iterator, Mapping
+
+from core.cost_receipts import (
+    BUCKETS,
+    RETRY_NONE,
+    RETRY_KINDS,
+    STAGES,
+    STAGE_BUCKET,
+    CallUsage,
+    CostReceipt,
+    CostReceiptLedger,
+    load_receipt_price_table,
+    make_call_id,
+)
+
+__all__ = [
+    "Attribution",
+    "CostRecorder",
+    "MeteredClient",
+    "extract_usage",
+    "open_cost_recorder",
+    "resolved_model_of",
+]
+
+
+@dataclass(frozen=True)
+class Attribution:
+    """Whose call this is: one task, one stage, one kind of retry."""
+
+    task_id: str
+    stage: str
+    retry_kind: str = RETRY_NONE
+    attempt_index: int = 0
+
+    def __post_init__(self) -> None:
+        if self.stage not in STAGES:
+            raise ValueError(f"unknown stage {self.stage!r}")
+        if self.retry_kind not in RETRY_KINDS:
+            raise ValueError(f"unknown retry kind {self.retry_kind!r}")
+        if not self.task_id:
+            raise ValueError("a metered call must belong to a task")
+
+    @property
+    def bucket(self) -> str:
+        return STAGE_BUCKET[self.stage]
+
+
+_CURRENT: ContextVar[Attribution | None] = ContextVar(
+    "cost_attribution", default=None
+)
+
+
+# ── Reading usage off whatever the provider returned ─────────────────────
+
+
+def _as_int(value: Any) -> int | None:
+    """Accept a token count only if it is really one.
+
+    ``bool`` is excluded on purpose — ``True`` is an ``int`` in Python and a
+    truthy flag arriving where a count belongs should not silently become one
+    token.
+    """
+    if value is None or isinstance(value, bool):
+        return None
+    if isinstance(value, int):
+        return value if value >= 0 else None
+    return None
+
+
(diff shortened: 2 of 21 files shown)


hyeonsangjeon:
**Held as draft on purpose — merge blocker is live, not hypothetical.**

`exp_gold_baseline` (`gold_ceiling_30_v2_sol_max.yaml`, `src_c33d9d55703fbf5d`) is mid-run right now:

- shard 0 → committed (`d9793ba`)
- shard 2 → committed (`b57e09e`)
- shard 1 → run [`33187134567`](https://github.com/hyeonsangjeon/gdpval-realworks/actions/runs/33187134567), still `in_progress`

The `exp_gold_baseline` final already on `main` is from an earlier run (`src_8513975c188f31a6`), not this one. This one has not merged.

Merging now breaks that merge after the spend, for the two reasons in the PR body (schema `1.3`→`1.4` identity check, and `grader_source_hash` covering all of `core/`). I reproduced the first one against the two real shard files now on `main`:

```
ShardMergeError: shard merge identity mismatch for schema_version:
s0=%271.3%27, current contract=%271.4%27
```

**Mark ready and merge once the merged `exp_gold_baseline` final for `src_c33d9d55703fbf5d` is on `main`.** No code change is needed to unblock — only the wait.

hyeonsangjeon:
**Merge blocker has cleared. Marked ready.**

The `exp_gold_baseline` run finished on its own while this PR sat in draft:

- shard 1 → [`33187134567`](https://github.com/hyeonsangjeon/gdpval-realworks/actions/runs/33187134567) completed/success (`ef03fdc`)
- 3-shard merge → `4dddf27`
- merged final for `src_c33d9d55703fbf5d` is on `main`, plus its auto-analysis (`d99ae50`)

Re-verified before flipping the flag:

- no grading or batch run is `in_progress` (the only live run is this PR's own Backend Tests)
- every `_shards/` directory on `main` has a matching final beside it — nothing else is mid-merge

Rebased onto `d99ae50` (clean; `main` had only added grade data). Full suite on the rebased head: **`5171 passed, 6 skipped, 45 deselected`, exit 0.**

The ⚠️ section in the PR body is left in place deliberately — it is still the correct standing rule for the *next* schema or `core/` change, and the reproduction in it is real.

PR #276, opened 2026-08-29
fix(grading): a video container is not proof of silence

hyeonsangjeon:
## What the smoke found

The paid audio smoke (run `33262668933`, gold task `75401f7c`) passed — and that is how it did its job. It came back with `perception_call_count: 0` on a `.mp4` deliverable and a score of **26.79%**.

Eleven criteria of the form *"the sound effect file `Mountain Audio - Electricity.mp3` is audible during the opening logos.mp4 shot"* were marked **fail, 0 of 2**. The evidence string each one cites is `"audio_tracks": 1` — the container's own count of the audio it was being told it did not have.

## Root cause

`has_audio_content` returned `False` for every video container:

```python
if kind == "audio":
    return True
if kind != "zip":      # <- .mp4 is kind "video"
    return False
```

`False` is not "unknown". It is a positive claim that the file was examined and carries no audio, and `resolve_runtime_routing` uses it to demote an AUDIO criterion to TEXT. `.mp4` is disjoint from `GRADER_AUDIO_EXTENSIONS` (`.aac .flac .m4a .mp3 .ogg .wav`), so the extension test and the probe agreed on the wrong answer.

This is the `.zip`-of-stems bug one container along, and the existing docstring already had the reasoning — *the container extension is a fact about packaging, not about the medium*. Video was simply never given the archive's treatment.

## Why `None` and not `False` on failure

Demotion is **not** exclusion. An excluded item leaves the denominator; a demoted one gets scored by a judge that cannot hear. So a probe that guesses `False` converts *"we did not listen"* into *"the work is bad"* — which is the contract the grader is supposed to hold. Missing decoder and unreadable container both answer `None`.

For contrast, the same task's three **visual** criteria behaved correctly: `required_visual_render_target_unavailable`, `score_excluded: true`, out of the denominator.

## Blast radius

**21 items over 2 tasks**, both Film and Video Editors. Re-measured against the Stage 3 corpus after #291 landed:

| task | recorded | items demoted | points not awarded |
|---|---|---|---|
| `75401f7c` | 26.34% | 14 | 18.0 / 56 |
| `e222075d` | 16.17% | 7 | 10.3 / 60 |

Small in aggregate, severe per task: crediting every demoted item in full moves the 174-task mean 80.19% → 80.48%, but those two tasks' recorded percentages are floors rather than measurements.

The one place it is *not* small is the sector table. Both tasks are Information, which #291 otherwise ranks last at 71.19%; without them it is **81.18%** and mid-table. That line was a fact about the grader on its way to being read as a fact about the sector.

*(An earlier revision of this section said 28.6 points. That total came from the superseded `src_ce7d4978bce9effc` grade files, where `e222075d` lost 10.6. Against the run Stage 3 actually closed on it is 28.3. The same two tasks, either way.)*

Criteria demoted against `.pdf` / `.xlsx` / `.docx` are the rule working as designed and are untouched.

## Tests

`tests/test_a_reel_with_a_soundtrack_gets_listened_to.py` — 13 tests, encoding real MP4s with PyAV (already pinned in `requirements.txt`), asserting on the three criterion strings copied verbatim from the smoke's graded output.

Covered: video with a soundtrack → `True`; genuinely silent video → `False` (the saving is preserved); audio-only `.mp4` → `True`; corrupt container → `None`; missing PyAV → `None`; end-to-end routing on the real criteria; non-regression on `.xlsx`, `.mp3`.

**Negative control:** reverting the branch fails 7 of 13, including all three end-to-end routing assertions.

## The fingerprint move, and why it is now expected

`core/tools/read_deliverable.py` is covered by the grader source hash, so this change moves the gold fingerprint. Recomputed on the rebased branch:

| | |
|---|---|
| gold_ceiling_185_v2_sol_max | `955be41edc4aff19` → `450aee03cf6cb840` |
| gold_smoke_audio_v2_sol_max | `e23d7898c04f6b63` → `25d79158aa1fa3b2` |

The original hold — eleven shards in flight at `955be41edc4aff19` — is discharged. Those shards landed, four paid attempts failed to finish the twelfth task, and #291 closed Stage 3 at **174/185, blocked rather than final**. The 174 and their eleven cost ledgers are frozen and preserved; nothing here rewrites them, and `stage3_partial_inventory.py` pins the old hash as a constant rather than recomputing it, so this merge cannot disturb that evidence.

So the fingerprint moving is no longer a cost — Stage 3 already needs a re-run at a new hash and a new run ordinal for reasons unrelated to this fix. `305-resume-granularity.md` recommends bundling further grader changes with exactly this kind of move, where their marginal cost is zero.

**Old and new results must never be mixed.** The 174 were measured at `955be41edc4aff19`; anything graded after this merge is a different measurement.

## Rebase and re-audit

Rebased from a 24-commit-stale base onto `bc806de`; the diff is unchanged (2 files, +258/−5) and both hash figures above re-verify at the new base.

Negative control re-run after the rebase: surgically excising only the `if kind == "video":` branch — 692 bytes, leaving the helper in place — fails **7 of 13**, including all three end-to-end routing assertions on criteria copied from the graded output. Restoring it passes 13 of 13.

**No paid re-grade follows this merge.** The 2-task audio smoke at the new fingerprint is prepared separately and is the owner's call to dispatch.

### DIFF
--- a/batch-runner/core/tools/read_deliverable.py
+++ b/batch-runner/core/tools/read_deliverable.py
@@ -1249,11 +1249,12 @@ def has_audio_content(path: Union[str, Path]) -> Optional[bool]:
     -- tempo, key, vocals, mix -- was graded end to end without a single
     listening call.
 
-    ``True`` for an audio file or an archive holding one. ``False`` is a
-    positive claim and means the file was examined and carries no audio.
-    ``None`` is an admission -- missing file, unreadable archive, or a member
-    list that was cut short and may have stopped one entry before the audio --
-    so that routing never promotes on a guess.
+    ``True`` for an audio file, an archive holding one, or a video whose
+    container carries an audio track. ``False`` is a positive claim and means
+    the file was examined and carries no audio. ``None`` is an admission --
+    missing file, unreadable archive, absent decoder, or a member list that was
+    cut short and may have stopped one entry before the audio -- so that
+    routing never promotes on a guess.
     """
     p = Path(path)
     if not p.is_file():
@@ -1261,6 +1262,18 @@ def has_audio_content(path: Union[str, Path]) -> Optional[bool]:
     kind = _kind_of(p)
     if kind == "audio":
         return True
+    if kind == "video":
+        # The same mistake as the ``.zip`` above, one container along. A reel
+        # is delivered as ``.mp4``; ``.mp4`` is disjoint from the audio
+        # extensions; so "the sound effect is audible during the opening
+        # shot" was demoted to TEXT and answered by a judge that had read the
+        # file's metadata and could not hear it. It scored zero -- not
+        # excluded, *failed* -- on evidence reading ``"audio_tracks": 1``.
+        #
+        # That count is the honest answer to this question and was already
+        # being computed one call away.
+        count = _video_audio_track_count(p)
+        return None if count is None else count > 0
     if kind != "zip":
         return False
     try:
@@ -2096,6 +2109,30 @@ def _op_probe_audio(p: Path, _scope: Dict[str, Any]) -> Dict[str, Any]:
     info = _probe_audio_impl(p, basic=False)
     info["kind"] = "audio"
     return info
+
+
+def _video_audio_track_count(p: Path) -> Optional[int]:
+    """How many audio streams a video container carries, or ``None``.
+
+    Deliberately not routed through ``_probe_video_impl``: that helper returns
+    early when a container has no *video* stream, which would report ``None``
+    for a file that is nothing but sound. Counting the audio streams directly
+    answers the question that was asked.
+    """
+    try:
+        import av  # type: ignore
+    except ImportError:
+        return None
+    try:
+        container = av.open(str(p))
+    except Exception:  # noqa: BLE001
+        return None
+    try:
+        return len(container.streams.audio)
+    except Exception:  # noqa: BLE001
+        return None
+    finally:
+        container.close()
 
 
 def _probe_video_impl(p: Path, basic: bool = False) -> Dict[str, Any]:
--- a/batch-runner/tests/test_a_reel_with_a_soundtrack_gets_listened_to.py
+++ b/batch-runner/tests/test_a_reel_with_a_soundtrack_gets_listened_to.py
@@ -1,0 +1,216 @@
+"""A criterion about sound must not be answered by a judge that cannot hear.
+
+``resolve_runtime_routing`` demotes an AUDIO criterion to TEXT when the files
+it is pointed at carry no audio, and that rule is right: "the mix is balanced"
+has no meaning against a spreadsheet, and sending it to a listening model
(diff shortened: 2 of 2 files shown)


hyeonsangjeon:
## CI evidence

Both runs are green on `fa1335c`.

| Where | Result |
|---|---|
| PR check (`pytest`) | pass, 11m16s — [run 33264098630](https://github.com/hyeonsangjeon/gdpval-realworks/actions/runs/33264098630) |
| Full backend suite, local py310 | `5843 passed, 6 skipped, 45 deselected in 686.67s`, exit 0 |

The baseline on merged main `7c92000` is `5830 passed, 6 skipped, 45 deselected`. The delta is **+13 passed and nothing else** — exactly the 13 tests this PR adds, with no existing test changing status.

### Negative control

The tests were checked against a deliberately broken fix rather than only against the working one. Patching the new branch to `if kind == "video" and False:` — i.e. restoring the defect while leaving every other line of the PR in place — gives **7 failed / 6 passed**, and the failures are the ones that matter:

- all three end-to-end `resolve_runtime_routing` assertions on the verbatim criteria from gold task `75401f7c`
- `test_a_video_with_a_soundtrack_reports_audio`
- `test_a_video_that_is_only_sound_reports_audio`
- both `None`-vs-`False` admission tests (unreadable container, absent decoder)

Restoring the line returns 13/13. So the suite fails for the defect this PR fixes, not merely for a syntax error somewhere near it.

Hold still applies — see the fingerprint table above. Nothing here changes that; this comment only records that the fix is regression-free and ready the moment the freeze lifts.

hyeonsangjeon:
## What this cost on the current gold Full

The Full now finishing at pre-fix source hash `955be41edc4aff19` gives the
defect a price. Searching every graded item in the ten complete shards for the
`audio_tracks` evidence string — the container's own count of the audio it was
declared not to have — returns **exactly two tasks**, both Film and Video
Editors. Every other task in the corpus returns zero:

| task | pct | perception calls | audio criteria | lost | rubric points lost |
|---|---|---|---|---|---|
| `75401f7c-396d-406d-b08e-938874ad1045` | 26.34% | **0** | 15 | 14 | 19.0 |
| `e222075d-5d62-4757-ae3c-e34b0846583b` | 16.17% | **0** | 8 | 8 | 12.3 |

`perception_call_count: 0` on both is the fix's whole justification. The
corpus made **580** perception calls overall, so perception is wired and
firing — these two deliverables were routed away from it because
`has_audio_content` returned `False` for an `.mp4`, and `False` is a positive
claim of silence rather than an admission of not having looked.

The two tasks scored 26.34% and 16.17% against a corpus mean near 80%. Both
were already named in `KNOWN_LIMIT_TASK_IDS`, so the loss was carved out of
the headline mean before the run rather than argued about after it — but
carved out is not the same as fixed, and 22 items across the two were scored
`fail` on criteria no model ever listened to.

Held deliberately: merging this changes the grader source hash, and the Full
is still mid-flight on shard 4. It goes in once that run is preserved.

PR #298, opened 2026-08-31
fix(grading): a refused audio call must not spend the task's listen

hyeonsangjeon:
## What the paid smoke found

Run `33363059548` routed all 21 damaged criteria to the listening model — PR #276 working as intended — and returned **no verdicts at all**. Two evidence strings came back:

```
audio_perception_failed:provider_error:BadRequestError   ×3 per task
audio_perception_failed:cap_exceeded                     everything after
```

Three defects sit behind that, and they are separable.

### 1. The cap was spent by calls that cost nothing

`AUDIO_CALL_CAP = 3` bounds what a task **spends**. `_calls_used` was incremented before the request and a request that raised never gave the slot back, so three rejections consumed a task's entire audio budget: `e222075d` had one of seven criteria attempted and six refused against a budget that had bought nothing.

The report is the worse half. Read off the artifact, `cap_exceeded` means *"we listened three times and stopped"* — someone tuning the cap upward would have been acting on a number that described nothing. What was true is that the model was never usable.

The refund alone would be a regression, so it lands with two counterweights:

- **a bounded failure budget** — otherwise a flaky endpoint turns three doomed requests into fourteen;
- **a short-circuit** for rejections that are deterministic for the task — the next criterion sends the same clip in the same shape and gets the same 400.

5xx and timeouts stay charged: they may have been billed for work nobody saw. Over-charging costs one criterion; under-charging costs money the ledger never learns about, and only one of those is recoverable.

### 2. `temperature` was the only sampling parameter anywhere on the grading path

Neither the main judge nor `VisionPerception` sends one. Audio was the sole exception and the sole caller whose every request came back 400.

A 400 body is not readable without buying another call, so the other candidate — request size — is ruled out on our side of the wire instead: `AUDIO_MAX_REQUEST_BYTES` refuses an oversized payload **by name**. The next smoke then has three distinguishable outcomes rather than one ambiguous one: a verdict (it was `temperature`), `audio_payload_too_large` (it was size), or a 400 still (it was neither, and the guessing stops).

### 3. The probe could not answer a question the file states plainly

Gold task `75401f7c` was marked **fail** on:

```
criterion:  The audio sample rate is 44.1 kHz or 48 kHz.
evidence:   "audio_tracks": 1, "video_tracks": 1
routing:    audio          perception_called: False
```

The routing was right and the listening path was down, so the judge fell back to the container probe — which, on an `.mp4`, answers with two track counts and nothing else. The rate is in the file. `ffprobe` prints it. `_probe_audio_impl` returns it. `_probe_video_impl` never looked.

This is the defect PR #276 fixed one field over, and fixing the routing while leaving it would have moved the bug into the next field rather than removing it. `_audio_stream_summary` now reports `sample_rate`, `channels` and `codec` under the same names the audio probe uses, and a **silent video reports `None`, not zeros** — a rate of zero is not a fact about any file.

## Negative controls

The refund is the risky edit, so the controls are where the review effort went:

| control | what it would catch |
|---|---|
| `test_successful_listens_still_stop_at_the_cap` | a refund written one line too broadly — a task with 14 audio criteria listening 14 times while its neighbour listens 3. `305` states that fairness invariant as B-1. |
| `test_a_failure_that_may_have_been_billed_stays_charged` (`None`/500/502/503/504) | widening the refund to failures that could have been billed |
| `test_a_recovered_endpoint_still_owes_its_earlier_failures_nothing` | refunded failures topping up the call cap |
| `test_the_size_limit_leaves_an_ordinary_clip_alone` | a guard set near real payloads, which would look exactly like the outage it rules out |
| `test_the_track_counts_are_still_there` | a probe change that fixes this criterion by breaking the one #276 fixed |
| `test_a_silent_video_reports_nothing_rather_than_zero` | absence rendered as a measurement |

The probe change was additionally mutation-tested five ways (drop the field, empty list → dict of `None`s, hard-coded 44100, one try/except around the whole dict, renamed key). Each was caught by the test written for it; none passed silently.

## Also: the vision render-target decision (no code change)

`tasks/rebuilding_grading_task/308-vision-render-target-decision.md`.

Measured rather than estimated, from the 174-task paid corpus: **9 of 7,326 items (0.12%)**, across 3 tasks — 4 on two `.mp4` deliverables, 5 on a `.zip` of React source. All 9 are `score_excluded`, **none is mis-scored**, and all 9 already bucket as `render_target_missing` (a harness limit) rather than `nothing_submitted` (a model fault).

No renderer is being built. A fixed frame cannot honestly answer *"the first non-logo content shot begins within 7.0 seconds of the start"* — frame 1 **is** the logo — and answering it anyway converts an honest exclusion into a confident wrong `fail`, which is the family of bug #276, the audio fix and the probe fix all exist to remove. The remaining limitation, and the one genuine classifier defect inside it, are written down with their measured size rather than left implicit.

## Test evidence

```
5922 passed, 9 skipped, 45 deselected, 2 warnings in 707.31s
```

Nothing in this PR calls a model or a network.


hyeonsangjeon:
## The 400 is decidable without buying another smoke

This PR's plan is to narrow three candidates — `temperature`, payload size, or "neither" — by dispatching a second paid smoke and reading which of the three outcomes comes back. That call is not needed. The cause is checkable statically, in the installed SDK, and it is none of the three:

**audio is not a member of the Responses content union.**

```
$ python -c "..."   # openai 2.46.0, the pinned version
Responses content union members:
  - ResponseInputTextParam
  - ResponseInputImageParam
  - ResponseInputFileParam
Chat Completions content union members:
  - ChatCompletionContentPartTextParam
  - ChatCompletionContentPartImageParam
  - ChatCompletionContentPartInputAudioParam        <-- the only one that exists
  - File
```

`ResponseInputContentParam = Union[ResponseInputTextParam, ResponseInputImageParam, ResponseInputFileParam]`. There is no audio member, in any spelling. A Responses request carrying audio is refused with a 400 before a model hears a second of it — regardless of `temperature`, and regardless of size. `ResponseInputAudioParam` does exist under `openai.types.responses`, which is what made this hard to see, but it belongs to the **Evals graders**, not to Responses message content. Reading the shape off that type is why the audio unit test went on passing while every real call failed.

So the three outcomes this PR is designed to distinguish collapse to one: a 400, still, from both remaining candidates and from a corrected `temperature`. Every audio call this pipeline has ever made was rejected for this reason. There is no successful audio call in any committed grade payload, on any run, ever.

### What this does not invalidate

The cap-refund half of this PR is right and I am not arguing with it. A request the provider refused before running the model did not spend anything, and `cap_exceeded` describing a task that never listened is exactly the misreport it is called here. The failure budget and the deterministic short-circuit are the correct counterweights. Keep all of it.

What changes is the reason it will matter: after the endpoint is corrected, the refund stops being the thing that unblocks audio and goes back to being what it should be — a guard for the flaky case.

### The fix, and where it is

I have the endpoint change plus a re-encode and a raised cap on `fix/audio-payload-and-cap`, with a generalised guard (`tests/test_audio_goes_to_the_endpoint_that_accepts_it.py`) that asserts union membership rather than asserting a key by hand, so the next same-named-type substitution fails in CI instead of in a paid run. Three parts:

1. `chat.completions.create` with `{"type": "input_audio", "input_audio": {"data": ..., "format": ...}}` and `modalities=["text"]`.
2. Re-encode the 30 s slice to 16 kHz mono. Measured on the two real deliverables: **7.68 MB → 1.28 MB** per call. That makes the size question moot rather than guessed at — worst-task traffic drops 246 MB → 41 MB.
3. `AUDIO_CALL_CAP` 3 → 32. Counted from the checked-in 220-task inventory: 11 tasks ask for more than three listening calls and the worst asks for 32, so the cap was starving them independently of the endpoint. 29 tasks have at least one audio criterion, 150 criteria in total, p95 = 14, median = 2.

### Suggested landing

Mine is currently based on `main` and touches the same three files, so it will conflict with this. I would rather not have two PRs fighting over `core/perception/audio.py`. Either:

- merge this one first and I rebase on top of it, which keeps the refund work intact and adds the endpoint fix as a follow-up; or
- I push the endpoint change onto this branch so it lands as one coherent PR.

Tell me which and I will do it. What should not happen either way is dispatching another paid smoke to distinguish `temperature` from size — both are already ruled out above, and the answer costs nothing to check.

PR #312, opened 2026-08-31
fix(envelope): record-only cost policy needs no account balance

hyeonsangjeon:
Stacked on #301 (`wt/silence-is-not-a-setting`). **Do not merge yet** — see the bottom.

## What was wrong

The free advance check read an account's remaining monthly credit out of the published plan, and a run under `record_cost_findings_only` could not pass without it.

Two things were wrong at once:

- **Nothing read it.** No code compared `available_monthly_credit_usd` against anything. It was carried on the result, printed in the summary, and that was the end of it.
- **The file is published.** A remaining balance is account information, not a setting for a run. Requiring it here meant somebody had to keep a balance current in a public file to be allowed to start a run.

## The two policies already existed

Nothing was invented. `block_on_cost_findings` and `record_cost_findings_only` are the only two policies in the code, and `block_on_cost_findings` is the one that actually stops a run on cost. So that is the only policy now required to name a per-run approved amount, and it fails closed without one.

## Reproducing the real path found three holes, not one

Before changing anything I fed the committed plan through the real `run_envelope_preflight` and varied one value at a time across 13 cases. No hand-made constants.

| | Situation | Before | After |
|---|---|---|---|
| **B** | record-only, credit key absent | 6 refusals (baseline is 5) — the assigned hole | **5**, no credit refusal |
| **G·K** | block policy, amount `null`/`-12`, assumptions unreadable so no ceiling | **the amount was never checked at all** | G: "no ceiling could be worked out…" · K: "must be greater than zero" |
| **J·L** | block policy, amount `.inf` / `.nan` | `.inf` passed silently (an unbounded budget approved); `.nan` raised an uncaught `InvalidOperation` | both refused: "is not a definite amount of money" |

**Root cause of G·K:** `check_cost_ceiling` sat inside `if conditions and loaded_catalog is not None`, so the amount was only ever judged on the way to comparing it against a worked-out ceiling. When the other half of the sum failed, the amount went unexamined — a plan saying "stop on cost" with no amount to stop at could reach a run place.

## What changed

| File | Change |
|---|---|
| `core/execution_envelope_cost.py` | `read_approved_maximum` — one place that turns a written amount into a number, so two callers cannot disagree about the same value. `.inf` and `.nan` refused by name. `check_approved_maximum` — faults in the amount alone, with no ceiling to compare against. |
| `core/execution_envelope_preflight.py` | Credit field deleted from the result, `as_dict()` and the printed summary; the key is now refused; the amount parse hoisted out of the ceiling branch; block-policy-with-no-ceiling now calls `check_approved_maximum`. |
| `experiments/execution_envelope/advance_check_plan.yaml` | Field removed, with a comment saying why it must not come back. |

Every existing refusal message is preserved verbatim. Exactly one new message was added.

## Two design decisions, stated so they can be overruled

1. **Under record-only, a volunteered `approved_maximum_usd` is still read** and still produces a ceiling **finding** — never a blocker. Dropping the read would destroy cost measurement, which is explicitly meant to continue. Both committed plans have it `null`, so nothing changes in practice.
2. **A leftover credit key is refused, not ignored.** A number sitting unread in a published file is an invitation to keep it up to date. One file carried it and it is removed in the same commit, so this is an explicit change rather than a silent one.

## Not in scope, and checked rather than assumed

`core/agentic_v2_stage_one_budget.py` demands an approved amount unconditionally. It looks like the same defect and is not: the module never mentions `policy`, and its plan declares no `policy:` key — its cost block says nothing has been approved for stage one, with the amount `null`. It is not under record-only, and its unconditional demand is correct fail-closed behaviour. Left untouched.

## An unpriced model stays unpriced

Refusing the amount does not turn a missing price into nothing owed. A regression test refuses the amount with `.inf` and then asserts `cost.unpriced_models` is non-empty and `cost.total_usd > 0`. Nothing is counted as zero dollars.

## No existing experiment changes meaning

The committed plan produces **the same five blocking problems before and after** — support, security and Model Wave, all pre-existing. No approval requirement comes back anywhere.

## Verification

| Check | Result |
|---|---|
| Real `run_envelope_preflight`, 13 cases | all as intended |
| Targeted 5 test files | 589 passed |
| Mutation proof (each new guard disabled in turn) | 5 / 4 / 5 failures, all restored |
| Full `pytest` | 6055 passed, 9 skipped |
| `mypy core/` | 170 errors in 32 files — baseline exactly; **0** in the two edited modules |
| `node --test` | 184 · 183 pass · 1 skip · 0 fail — baseline exactly |
| `npm run build` | clean |
| Model calls · grading runs · approved amounts · Azure logins | 0 · 0 · none · none |

## Before merging

- **The grader source fingerprint moves for all four configs.** Measured with the real `compute_grader_source_hash`; the pre-change `default_v2` value matches the one recorded on #301, so the stack's chain is intact.
- **`main`'s monthly-credit guard test currently fails against this branch head.** All four lines it flags are pre-#305 text inherited from the branch point. A 3-way merge against current `main` clears every one of them — 0 conflicts, 0 occurrences, 0 leftover fields, in both the plan and the spec. This goes green once `main` is pulled up the stack, which is tracked separately along with recomputing every fingerprint on current `main`.

Korean write-up: spec §13.33.


hyeonsangjeon:
**#336 으로 대체되었으므로 닫습니다.**

왜 개별 병합이 아니라 통합이었는지:

---

PR #313, opened 2026-08-31
fix(cost): report unpriced failed-task cost as no record, not $0.0000

hyeonsangjeon:
Stacked on #312 (`wt/record-only-needs-no-balance`). **Do not merge** — see the freeze note at the bottom.

## What was wrong

```
| Failed tasks | 2 ($0.0000) |
```

That reads as *"two failures, and they were free."* It also gets printed when the truth is *"two failures, and we have no idea what they cost."*

`summarize_cost_receipts` increments `failed_count` for **every** failed row carrying a receipt, but only adds to `failed_amount` when `_receipt_amount` returns a number. A failure billed against a model the price table has no entry for adds nothing, and the sum it never joined stays `0.0` — byte-identical to what a failure that made no model call leaves behind. The amount alone cannot separate them.

The fix publishes `failed_measured_tasks` beside the amount and qualifies the rendered row by it.

| case | before | after |
|---|---|---|
| no failures | `0 ($0.0000)` | `0 ($0.0000)` — unchanged |
| all priced | `1 ($0.4000)` | `1 ($0.4000)` — unchanged |
| **none priced** | `2 ($0.0000)` | **`2 (no record)`** |
| **partly priced** | `2 ($0.4000)` | **`2 ($0.4000, 1 / 2 priced)`** |
| genuinely free | `1 ($0.0000)` | `1 ($0.0000)` — unchanged |
| **receipts unreadable** | `2 ($0.0000)` | **`2 (no record)`** |
| **never ran** | `1건 · $0.0000` | **`1건 · 미확정`** |

`CostReceipt.free()` — *"The one honest `$0`. Reserved for paths that reach a verdict by rule rather than by asking a model."* — keeps printing `$0.0000`. That is the case the fix is careful not to break.

## Three paths, one defect

The Python and JavaScript summarisers are the same logic in two languages. Fixing one would have left the report and the dashboard saying different things about the same run, so all three move together.

| path | producer | renderer |
|---|---|---|
| report | `core/cost_projection.py` | `step6_report.py::_failed_task_cost` |
| dashboard aggregate | `scripts/cost-receipt.mjs` | — |
| dashboard UI | — | `src/lib/cost.ts::failedTaskCostCell` |

The UI reuses the existing `recorded` / `floor` / `unpriced` states. No new state was invented, and the cell keeps its amber styling — only the text, the tooltip, and a new `data-cost-state` attribute differ.

## Scope widened once, on the producer's own evidence

`not_run` receipts were also rendering `$0.0000`. Whether to count them as free was the one real judgement call here, and the code that creates them settles it:

```python
def not_run(cls) -> "CostReceipt":
    """This pipeline did not run. Not free — it did not happen."""
```

The producer refuses to call that state free; the cell was calling it free. `not_run` now counts as unpriced.

This changes one existing browser scenario's cell from `1건 · $0.0000` to `1건 · 미확정`. That run was **already** flagged as only partly priced, and the change removes a false "free" claim rather than adding a claim — the same class of rendering change accepted in #299.

## Backward compatibility

`failed_measured_tasks` is optional in the TS type. `grading_cost` is recomputed by `aggregate-grades.mjs`, so it gains the field automatically; `problem_solving_cost` is read from already-published `report_data.json`, which does not have it. When absent, the amount is trusted only if the whole run was priced — a fully priced run cannot contain an unpriced failure — and is otherwise treated as unmeasured. `step6_report._failed_task_cost` applies the identical fallback, and there is a test that strips the field from a real summary to prove both branches.

## Deliberately not fixed here

A failed task carrying **no receipt at all** is skipped before `failed_count` is reached, so a run with one failure renders `Coverage 1 / 2 tasks (50.0%)` next to `Failed tasks | 0`. That is the **count**, not the amount, and `failed_task_count` already has published JS/TS consumers — changing what it means is a wider blast radius than changing what is printed beside it. Filed separately with the reproduction.

## Verification

Reproduced first on the real chain, with no hand-written receipt JSON: `build_receipt(...)` → step2 export row → `_compute_cost_summaries` → `_build_markdown`, then read the rendered line. Seven cases, all as intended.

| | |
|---|---|
| full `pytest` | **6066 passed · 9 skipped · 45 deselected** — baseline 6055 **plus exactly the 11 tests added here** |
| `mypy core/ --ignore-missing-imports` | 170 errors in 32 files — **baseline matched** |
| `node --test` | 186 pass · 1 skip · **0 fail** (3 added) |
| real-browser cost test | passes |
| model calls · grading runs · approved amounts · Azure logins | **0 · 0 · none · none** |

Four mutants, all killed:

| | mutation | result |
|---|---|---|
| M1 | Python producer counts unpriced failures as measured | 6 tests failed |
| M2 | Python renderer always prints the amount | 4 tests failed |
| M3 | JS producer, same mutation as M1 | 3 tests failed |
| M4 | UI always prints the amount | **real browser failed** |

M4 is the defect reproduced end-to-end through the built bundle:

```
+ actual   '1건 · $0.0000'
- expected '1건 · 미확정'
    at assertRanButUnpriced (scripts/__tests__/cost-receipts.browser.mjs:637)
```

All four were restored, no test markers remain, and the build was re-run clean afterwards.

## Checked rather than assumed

`core/execution_envelope_preflight.py:1664` (`if loaded_catalog is not None:`) looks like the fail-open shape fixed in #312, but `load_task_catalog` either raises `ValueError` (6 sites) or returns a catalog — it has no path that returns `None`. Already fail-closed; not touched, not duplicated.

## Grader fingerprint — moves

`core/cost_projection.py` is a hash input. Computed with the real `compute_grader_source_hash`.

| config | before (`d32b1bb`) | after |
|---|---|---|
| `default_v2` | `d29dfe3acabdb048` | **`fcfa3e83f444c10d`** |
| `gold_smoke_audio_v2_sol_max` | `f5fdb042cf15a29e` | **`7bbda955b35834a5`** |
| `gold_ceiling_185_v2_sol_max` | `f21cd8888ae40b7b` | **`bfb37656f9448b79`** |
| `default_v2_sol_max` | `b9d618eaad290680` | **`9d33d5db7e3542c9`** |

All four "before" values match what #312 recorded, so the stack is unbroken.

## Freeze

**#292 → #294 → #295 → #297 → #299 → #301 → #312 → this PR are all held.** Nothing merges until the fingerprints are recomputed on current `main` and each stacked head.

One thing a reviewer will notice: this branch point **predates #305**, so its copy of the spec still carries text #305 removed, and `main`'s guard test is absent here. That was left alone on purpose — hand-editing the exact lines #305 changed would only create a conflict. It resolves when `main` is merged up the stack.

Spec: 13.34.


hyeonsangjeon:
**#336 으로 대체되었으므로 닫습니다.**

왜 개별 병합이 아니라 통합이었는지:

---

PR #314, opened 2026-08-31
docs(qa): describe max_retries as answers in all, not extra tries

hyeonsangjeon:
Stacked on #313 (`wt/failed-cost-is-not-zero`). **Not for merge** — see the bottom.

## What is wrong

`qa.max_retries` is named as a count of retries. The loop spends it as a budget of answers.

`_run_task_with_qa` raises `qa_attempts` once per answer it has *reviewed and found wanting*, then breaks the moment that count reaches `qa_max_retries`. So the number written is the number of answers produced, and the number of replacements it buys is one less.

Measured by running the shipped `run_inference` and counting model calls — not by reading the code and doing the arithmetic by hand:

| written | answers | reviews | banner | **replacements** |
|---|---|---|---|---|
| absent (default 2) | 2 | 2 | `2/3` | 1 |
| `0` | 1 | 1 | — | 0 |
| **`1`** | **1** | **1** | — | **0** |
| `2` | 2 | 2 | `2/3` | 1 |
| `3` | 3 | 3 | `2/4`, `3/4` | 2 |

`0` and `1` are indistinguishable runs.

## Why that matters

Parsing every experiment YAML through the loader rather than grepping:

| written | self-review blocks |
|---|---|
| **`1`** | **29** |
| `2` | 1 (`exp997`) |
| `3` | 2 (`exp998`, `exp999`) |

Twenty-nine of thirty-two are written as 1. Eleven of them (`exp001`–`exp011`) carry the inline comment `# max regeneration attempts on QA failure`. None has ever regenerated anything.

A bare `grep 'max_retries: 1$'` reports 18, because eleven lines carry a trailing comment. The 29 is the parsed figure.

## Why the behaviour is not changed

Making the code match the name is one character: `>=` → `>`. That would hand each of those twenty-nine experiments one more paid answer than the run they are a record of, changing the conditions of work already done.

So nothing a run does is touched. Four descriptions are corrected instead:

| where | was | now |
|---|---|---|
| progress banner (`step2_run_inference.py`) | `QA attempt 2/3` while producing the second and last answer | `QA attempt 2/2` |
| run-place comparison (`execution_envelope_preflight.py`) | "how many times it may answer again" | "how many answers it may produce in all" |
| `QAConfig` default (`experiment_config.py`) | `# Number of retries on QA failure` | what the number buys |
| ceiling docstring (`execution_envelope_cost.py`) | silent on the over-count | why it is deliberate |
| `README.md`, `README_KR.md` | bare `max_retries: 3` | `3 → 2 replacements, 1 → none` |

The twenty-nine YAMLs are left alone deliberately: they are records of runs already performed. Only the forward-facing README templates are corrected.

## The ceiling over-count is left in place, on purpose

`max_attempt_counts` bills `N + 1` answers where the loop produces at most `N`. That is the safe direction — a ceiling that is too high can only refuse a run that would have fit; one that is too low waves through a run that will not. The arithmetic is unchanged. The docstring now explains why, and a test pins it, so the next person to spot the discrepancy has to make a decision rather than perform a tidy-up.

## Tests

15 new tests in `test_self_qa_attempt_budget.py`. They drive the real `run_inference` with the executor and reviewer stubbed, and count calls. The built-in default is read *through the loader* by omitting the key, so a change to either the dataclass or the parse shows up.

Mutation-checked four ways:

| mutant | result |
|---|---|
| banner denominator back to `+1` | **2 fail** |
| **`>=` → `>` (the forbidden change)** | **13 fail** — 7 new + 2 pre-existing in `test_silent_corruption_fixes.py` |
| comparison description reverted | **1 fail** |
| ceiling lowered by one | **3 fail** — 1 new + 2 pre-existing |

The second row is the point: anyone "fixing" the off-by-one by changing behaviour turns 13 tests red immediately. Two mutants also killed pre-existing tests, which says the new ones point at the same facts the existing safety net already holds.

## Already correct, not duplicated

- `execution_envelope_preflight.py:1455` already names its own variable `written_qa_attempts` — the comparison logic was already thinking in answers; only the printed description was wrong.
- `execution_environment_readiness.py:1154-1166` already says "attempts", which is the right word under the real semantics, and its `<= 0` rejection is already fail-closed. Untouched.

## Verification

| | |
|---|---|
| `pytest` | **6081 passed** · 9 skipped · 45 deselected — baseline 6066 + 15 |
| `mypy core/` | 170 errors in 32 files — unchanged |
| `node --test` | 187 · 186 pass · 1 skip · **0 fail** |
| `tsc --noEmit` | clean |
| `npm run build` | clean |
| browser test | exit 0 |
| model calls · grading runs · approved amounts | **0 · 0 · none** |

## Grader fingerprints

Three hashed `core/` files are touched, so all four move. Computed with the real `compute_grader_source_hash`; the before column was recomputed from a clean checkout of the branch point rather than quoted from memory.

| config | before (`b3e6789`) | after |
|---|---|---|
| `default_v2` | `fcfa3e83f444c10d` | **`d9c1b34d915212b6`** |
| `gold_smoke_audio_v2_sol_max` | `7bbda955b35834a5` | **`479e507c39210d8e`** |
| `gold_ceiling_185_v2_sol_max` | `bfb37656f9448b79` | **`9cbf308d7ea7d637`** |
| `default_v2_sol_max` | `9d33d5db7e3542c9` | **`d9b4abcdbcccf914`** |

All four before-values match #313's after-values, so the stack is unbroken.

## Do not merge

#292 · #294 · #295 · #297 · #299 · #301 · #312 · #313 and this one are all held. Nothing merges until every fingerprint is recomputed on current `main` with the real hash function, after the paid run in flight has pinned its evidence.


hyeonsangjeon:
**#336 으로 대체되었으므로 닫습니다.**

왜 개별 병합이 아니라 통합이었는지:

---

PR #315, opened 2026-08-31
fix(cost): a task with no receipt is a hole in the record, not a zero

hyeonsangjeon:
Stacked on #314 (`wt/qa-attempts-are-not-retries`). **Do not merge.** The whole chain — #292 → #294 → #295 → #297 → #299 → #301 → #312 → #313 → #314 → this — is held until every fingerprint is recomputed against current `main`.

## The defect

`summarize_cost_receipts` walked the run's rows and dropped every row that carried no receipt for the field being summarised. The dropped row still counted in `total_tasks` and in the coverage denominator, so the run knew perfectly well it had happened — it just stopped counting it everywhere else. One `continue`, three wrong figures:

```
| Coverage       | 1 / 2 tasks (50.0%) |   ← one task's cost is unknown
| Receipt status | complete            |   ← claims the record is whole
| Total          | $0.4200             |   ← labelled "Total", over half a run
| Failed tasks   | 0 ($0.0000)         |   ← the failed task does not exist
```

1. `failed_task_count` lost the failure.
2. Completeness was judged as `counts["complete"] == len(receipts)` — which asks only whether the *kept* receipts agree, a question the discarded rows can never answer. `complete` then flips the report header from `Recorded so far` to `Total` and populates `estimated_cost_usd`.
3. `cost_per_successful_deliverable_usd` rides on the same flag, so an understated per-unit figure was published as settled.

## Why a receiptless row is routine, not hypothetical

- `core/result_projection.py:57-59` **drops the key** rather than writing a null, deliberately, so a pre-instrumentation run does not gain a field that reads like a recorded zero.
- Resume re-uses `prevalidated_progress["results"]` — rows written by an earlier build.
- `costReceiptsByTask` in `scripts/aggregate-grades.mjs:359` only records tasks whose `grading_cost` projects non-null, then `gradingCostSummary` maps over **every** graded task with `receipt: receipts.get(...) ?? null`. A graded run where some tasks carry a cost and others do not is the normal shape on the dashboard side.

## After

| | before | after |
|---|---|---|
| `failed_task_count` | `0` | **`1`** |
| `status` | `complete` | **`partial`** |
| `estimated_cost_usd` | `0.42` | **`None`** |
| report header | `Total \| $0.4200` | **`Recorded so far \| $0.4200`** |
| failed row | `0 ($0.0000)` | **`1 (no record)`** |
| per deliverable | a number | **`no record`** |

`known_cost_usd` is unchanged. What *was* recorded is still published — as a floor, not as an answer.

## Scope

**Only `complete` moves.** A run already reading `partial` / `unavailable` / `not_run` is already not claiming to be whole. A run where every row carries a receipt never reaches the new branch. A run where *no* row carries one still returns `None`, so pre-instrumentation experiments still read "no record". No fully-recorded experiment changes what it says.

Deliberately **not** done here:

- No rendering change. `summaryTotalCell` in `src/lib/cost.ts:346-377` already had the `≥ …` floor path with its Korean tooltip; the summariser was simply never routing to it.
- No change to `step6_report._failed_task_cost` — #96 already made it print `N (no record)` and `N ($X, M / N priced)`. It only needed honest inputs.
- No new `missing_reasons` value; that list is shared by the schema, the producer and the dashboard label map.
- The Python status chain is **not** restructured into the JS shape — that is a separate item, and folding it in here would change the verdict for hole-free runs too.

## The browser test's fixture already contained the defect

`scripts/__tests__/cost-receipts.browser.mjs` renders the real built dashboard and reads the text off the page. Its `fullyPricedRows` scenario has a task that was graded while the judge recorded no cost — deliberately, since that is the only way to test 기록 없음 against 미채점. So the grading side is 3 rows / 2 receipts.

The page printed `$0.1300` as the grading **total**, immediately beside a cell saying that third task's cost was unknown. It now reads `≥ $0.1300` with the "2 of 3 priced, so this is a floor" tooltip. The problem-solving half of the same payload has a receipt on every row and is untouched — still `$0.3400`, still `recorded`. One payload, two fields, only the one with a hole in it moves.

## Verification

| | |
|---|---|
| full `pytest` | **6101 passed, 9 skipped, 45 deselected** (baseline 6081 + the 20 new tests) |
| `mypy core/ --ignore-missing-imports` | 170 errors in 32 files — **baseline exactly** |
| `node --test 'scripts/__tests__/*.test.mjs'` | **193 · 192 pass · 1 skip · 0 fail** (baseline 187 + the 6 new) |
| `tsc --noEmit` | clean |
| `npm run build` | clean |
| `cost-receipts.browser.mjs` | exit 0 |

Everything is driven through production code: `build_receipt` for receipts, `project_result_row` to demonstrate the dropped key, `build_cost_summaries` for the summary, `step6_report._build_markdown` for the table, and the real page for the dashboard. No hand-made constants.

### Mutation proof — 10 applied, 10 caught, 0 survivors

| # | mutation | |
|---|---|---|
| M1 / M6 | the hole no longer counts its failure | caught |
| M2 / M7 | the downgrade is removed (`complete` allowed again) | caught |
| M3 / M8 | the downgrade fires on *every* status, not only `complete` | caught |
| M4 / M9 | every hole counts as a failure, success or not | caught |
| M5 / M10 | the hole stops being counted at all (the original defect) | caught |

M1–M5 Python, M6–M10 JavaScript — the two summarisers mirror each other, so both need proving. M3/M8 and M4/M9 are the negative controls: one guards against widening the fix, the other against confusing "we have no cost for this task" with "this task failed".

Additionally, mutating the JS downgrade and rebuilding turned the **browser** test red, and restoring it turned it green — the fix is visible on the rendered page, not just in the summariser.

## Grader fingerprints

`core/cost_projection.py` is inside the hashed set, so all four move. Computed with the real `compute_grader_source_hash`; the before column was re-derived from a detached checkout of `4b89db0`, and matches #314's recorded after-values exactly, so the stack is unbroken.

| config | before (`4b89db0`) | after |
|---|---|---|
| `default_v2` | `d9c1b34d915212b6` | `08c0458412ba286d` |
| `gold_smoke_audio_v2_sol_max` | `479e507c39210d8e` | `1728f26053b9aaea` |
| `gold_ceiling_185_v2_sol_max` | `9cbf308d7ea7d637` | `917665a7ba4f8325` |
| `default_v2_sol_max` | `d9b4abcdbcccf914` | `7728067635a3f59d` |

`scripts/cost-receipt.mjs` and the test files are outside the hashed set.

No model calls, no grading runs, no approval amounts, no cloud logins were involved in producing this change.


hyeonsangjeon:
**#336 으로 대체되었으므로 닫습니다.**

왜 개별 병합이 아니라 통합이었는지:

---

PR #317, opened 2026-08-31
fix(cost): a receipt names the table that priced it, not the one open now

hyeonsangjeon:
Stacked on #315. **Do not merge** — the whole chain is held until the fingerprints are recomputed on current `main`.

## What was wrong

The price-table fingerprint on a receipt is a claim a reader can act on: fetch those exact bytes, re-price the usage, get the amount back. `load_receipt_price_table` states the purpose in as many words — *"a receipt that records it can be checked against the exact bytes that priced it."*

The ledger was not recording that. Two places collapsed several tables into one:

| where | what it did |
|---|---|
| `receipt_for` → `build_receipt` | stamped whichever table **this process** opened the ledger with, over rows an earlier process had already settled under an earlier price file |
| `summarise_receipts` | `sha = sha or receipt.price_table_sha256` — first non-null wins, the rest discarded |

There is no check at ledger-open either: opening a ledger whose rows were priced by another table raises nothing.

## Reproduced through the real ledger

One ledger file, two processes, two price tables — a resume, or a shard started after a rate review. Real `CostReceiptLedger`, real `load_receipt_price_table`, real `reserve`/`settle`. No hand-made receipts.

January charges $10/$20 per million; February charges $100/$200.

```
rows actually stored:   t-jan  cost=30   sha=0d968779…
                        t-feb  cost=300  sha=d84e4aa8…

before  receipts:  t-jan $30 under d84e4aa8    t-feb $300 under d84e4aa8
        roll-up:   complete $330 under d84e4aa8

after   receipts:  t-jan $30 under 0d968779    t-feb $300 under d84e4aa8
        roll-up:   complete $330 sha=None
```

$30 is not what `d84e4aa8` charges for that usage — it charges $300. The receipt pointed at bytes that disprove its own amount, and a reader who checked had no way to tell whether the money or the provenance was wrong. Which table got named also flipped with task ordering: the same run summarised in the other direction gave the other answer, both stated with equal confidence.

## The fix

Ask the rows first; use the caller's table only where they say nothing.

| rows name | receipt records |
|---|---|
| one table | that table |
| two or more | **nothing** — any single answer is contradicted by the money beside it |
| nothing at all | the caller's table (unchanged) |

The third row matters: ledgers written before the column was populated, and ledgers opened with no price list, leave it null on every row. That is silence, not disagreement — blanking it would lose provenance that older runs still have.

## Not a new rule — the other half already did this

- `core/cost_projection.py:600` — `price_tables[0] if len(price_tables) == 1 else None`
- `scripts/cost-receipt.mjs:529` — distinct set, then `length === 1 ? [0] : null`

Both already refused to guess. The ledger was the one layer still picking, so the two layers disagreed about the same run. **Neither file is touched here; no JavaScript or TypeScript changes at all.**

## Deliberately unchanged

- **No amount moves.** Each figure was correct under the table in force when it settled. The doubt is about provenance, not money.
- **No status changes.** Adding a `missing_reason` would flip `complete` → `partial`, retro-downgrading published runs whose amounts are fine. `build_receipt`'s status is `PARTIAL if reasons else COMPLETE`, so a new reason cannot be added without that effect.
- **A single-table run reports exactly what it did before** — the rows agree, so the first branch returns the same value.

## Verification

| check | result |
|---|---|
| new tests | 19, through the real `reserve`/`settle` path |
| mutation | **10/10 caught, 0 survivors**; source restored byte-identical |
| full `pytest` | **6120 passed, 9 skipped, 45 deselected** (baseline 6101 + 19) |
| `mypy core/` | 170 errors / 32 files — baseline, unchanged |
| `node --test` | 193 · 192 pass · 1 skip · 0 fail — unchanged (no JS change) |
| browser test | **could not run** — `@playwright/test` is declared in `package.json` but absent from the shared install on this machine. Zero JS/TS bytes changed, so there is no path by which the rendered page could differ. Recording the gap rather than claiming a pass. |
| model calls · grading runs · approval amounts · Azure logins | 0 · 0 · none · none |

Two negative controls in the mutation set: one checks that fixing this does not blank the fingerprint on rows that simply predate the column, the other that an **abandoned** call — never sent, so never priced by anything — cannot manufacture a disagreement.

## Grader fingerprint

`core/cost_receipts.py` is hashed, so all four move. Computed with the real `compute_grader_source_hash`; the before column was re-derived from `cadbad2` with the same function and matches #315's after column exactly, so the stack is unbroken.

| config | before (`cadbad2`) | after |
|---|---|---|
| `default_v2` | `08c0458412ba286d` | `7a280ebd8c59451a` |
| `gold_smoke_audio_v2_sol_max` | `1728f26053b9aaea` | `1602fa7ec7be87de` |
| `gold_ceiling_185_v2_sol_max` | `917665a7ba4f8325` | `8f83c5ef148a7f1c` |
| `default_v2_sol_max` | `7728067635a3f59d` | `2a06ba6dec103b3a` |

#292 → #294 → #295 → #297 → #299 → #301 → #312 → #313 → #314 → #315 → this one are all held. Nothing merges until every fingerprint is recomputed on current `main`.


hyeonsangjeon:
**#336 으로 대체되었으므로 닫습니다.**

왜 개별 병합이 아니라 통합이었는지:

---

PR #318, opened 2026-08-31
fix(grading): a row that never answered has not answered yes

hyeonsangjeon:
Stacked on #317 (`wt/a-receipt-names-the-table-that-priced-it`). **Do not merge** — the twelve-PR stack is held pending a fresh fingerprint preflight on current `main`.

## The defect

`summary.cost.usage_complete` is the run's statement that the token counts behind its cost total all arrived. The fold in `_compute_summary` read a **missing** per-task flag as `True`:

```python
usage_complete = usage_complete and bool(task.get("usage_complete", True))
```

Two failures compound in the same loop. The counters immediately above default an absent token count to `0`, so a silent row makes the total **smaller** — and then that same row's silence certified the total **complete**. A shrunken number stamped whole.

Sixty lines away in the same file, the warning at the sector rollup reads the identical key strictly (`row.get("usage_complete") is not True`) and prints *"this run's cost total is no longer complete"*, with a comment naming the fold as the thing that carries the doubt. The two lines disagreed about the same key.

Every other reader already agreed with the warning, not the fold:

| reader | absent means |
|---|---|
| `step8_grade._compute_summary` | **complete** ← the bug |
| `step8_grade` sector-rollup warning | not complete |
| `core/tool_calling_judge.py` ×3 | not complete |
| `step2_run_inference.py` ×2 | not complete |
| `step6_report.py` | not complete |
| `core/agentic_endpoints.py` ×2 | not complete |
| `step9_merge_shards.py` | **raises** unless boolean |

## Reachability

Both doors into the fold feed it rows read off disk: resume seeds it with `list(existing_tasks)`, and `step9_merge_shards` **imports this very function**. A freshly graded row always carries the key out of `TaskGrade`, so only a loaded row can be silent.

`_validate_grade_task_set` was checking the wrong field. Its comment states the goal exactly — *"the flag must be present and must be a boolean, so the fold that produces the run's total can never be handed a missing or truthy-ish value"* — but it validated only `summary.cost.usage_complete`, a value `_compute_summary` **discards and recomputes from the rows**. A payload tidy at the top and silent underneath passed.

## Reproduced on a published run

One published shard of a real graded run, through the real `_compute_summary`:

| first row says | before | after |
|---|---|---|
| `True` (control) | complete | complete |
| `False` (control) | not complete | not complete |
| **nothing** | **complete** | **not complete** |

The last two differ only in whether the row spoke, and the published claim flipped between them. Money, tokens, latency, scores and sector rates are **identical** across all three — asserted as a whole-dict equality, so this is a claim fix, not a repricing.

## Blast radius: measured, and kept as a test

Every payload under `data/grades` folded both ways: **75 with rows · 62 unchanged · 13 change answer · 0 reachable.** All 13 are pre-1.2 files with no aggregate flag at all, and both doors demand a boolean there before reading a row. A row that says `False` is not silent — it answered — so an honestly incomplete run is on the unchanged side. The measurement is a test, so weakening the rule fails immediately.

## What is deliberately not changed

- **The schema.** `usage_complete` is optional because `schema_version` spans 1.0–1.4 and 1.0 predates the field; real 1.0 payloads sit on disk today. Requiring it would retroactively invalidate them. Silence is legal to *write* — reading it as a promise was the error.
- **`step9_merge_shards.py`.** It imports the same function, so it inherits the fix; its existing warning fires on honest degradation. Zero lines.
- **The `0` token defaults.** Once the flag is honest the total is a floor *and says so* — same resolution as the #103 family.
- **Frontend.** Zero bytes. The dashboard already renders this field.

The fold **degrades** rather than raises because it also runs on the checkpoint path, where raising would destroy already-graded work — the opposite of the module's stated goal. The resume guard **raises** because at that point the chunk has graded nothing, so refusing costs nothing and can name the faulty task id.

## The test harness had been exercising the defect

The full suite surfaced one failure — in a pre-existing test's own precondition. `tests/test_step9_merge_shards.py::_normalised_tasks` builds fixtures from a real **schema-1.0** corpus whose rows predate the field, stamps them as current-schema, and folds them. Under the old rule that read as complete, so `test_usage_complete_true_when_all_shards_true` had been passing **without a single row ever having said so**.

The fixture now sets the flag the way today's producer does, beside the existing `grading_cost = CostReceipt.unavailable()` line, with the reason the two migrate in opposite directions written down. **Not one assertion in that file changed.**

## Verification

| | |
|---|---|
| `pytest` (2nd pass) | **6148 passed · 6 skipped · 45 deselected · 0 failed** (12m16s) — includes the 25 new tests |
| `pytest` (1st pass) | 1 failure, diagnosed above, fixed in the fixture |
| new test file alone | 25 passed |
| mutation battery | **10 mutations, 10 caught, 0 survivors** (M4 and M9 are the over-blocking negative controls) |
| `mypy core/` | 170 errors / 32 files — **baseline unchanged** |
| `node --test` | 193: 192 pass · 1 skip · 0 fail — baseline, no JS changed |
| real-browser test | **not run** — `@playwright/test` is absent from this machine's install. Recorded, not claimed. |
| model calls · grading runs · approval amounts · Azure logins | 0 · 0 · none · none |

Skip counts on this machine move by a few between runs with Docker/seccomp availability (the sandbox-security gates); this change touches no Docker path and the collected total was identical across both passes.

## Grader fingerprint

`step8_grade.py` is hashed, so the fingerprint moves. Computed with the real `compute_grader_source_hash`, before-column re-derived by checking out `040f383` separately:

| config | before (`040f383`) | after |
|---|---|---|
| `default_v2` | `7a280ebd8c59451a` | `461dc8dea13fb3aa` |
| `gold_smoke_audio_v2_sol_max` | `1602fa7ec7be87de` | `31752df82f934f52` |
| `gold_ceiling_185_v2_sol_max` | `8f83c5ef148a7f1c` | `a85c48d8765b2b95` |
| `default_v2_sol_max` | `2a06ba6dec103b3a` | `7ab61eaa96697e7f` |

All four "before" values match #317's recorded "after" values exactly, so the stack is unbroken. Test files are not hashed.


hyeonsangjeon:
**#336 으로 대체되었으므로 닫습니다.**

왜 개별 병합이 아니라 통합이었는지:

---

PR #319, opened 2026-08-31
fix(cost): a run that measured nothing is not a run that did nothing

hyeonsangjeon:
Stacked on #318. **Do not merge** — see the freeze note at the bottom.

## The defect

`summarize_cost_receipts` read the run's status off `amounts`, the list of per-task figures that survived `_measured_amount`:

```python
complete_run = counts["complete"] == len(receipts)
if complete_run:
    status = "complete"
elif amounts:                      # (1)
    status = "partial"
elif counts["unavailable"]:
    status = "unavailable"
else:
    status = "not_run"
```

`amounts` is empty whenever every measurable receipt sits at a $0 floor. `_measured_amount`, sixty lines above, nulls that floor on purpose and says why: *"under `partial` a zero means nothing was confirmed yet, which is absence, not a floor of zero."* It is the ordinary case, not an edge one — any run whose calls went to a model the price table has no entry for lands there, and `project_cost_receipt` raises if a non-complete receipt carries an estimate, so there is nothing else for `_receipt_amount` to return.

So **(1)** let a run of real paid calls, with real tokens on every one of them, fall past `partial` all the way to `not_run` — the status whose own constructor says *"This pipeline did not run. Not free — it did not happen."*

The same expression carried a second defect. **(2)** `counts["complete"] == len(receipts)` counted a task that genuinely never ran *against* the run, so two priced tasks beside one idle task withheld `estimated_cost_usd` although every receipt under it was whole. Work that did not happen is not a hole in the record: it contributed nothing, so it neither drags the status down nor stops a total being a total.

Reproduced through the real producer — a real `CostReceiptLedger`, priced by a real `load_receipt_price_table`, projected by the real `project_cost_receipt`:

```
A. every call went to a model the price table does not carry
   per-task statuses : ['partial', 'partial']
   RUN-LEVEL status  : not_run     <-- 2 tasks, 2 real model calls, 4M tokens

B. one task never ran, the rest priced cleanly
   per-task statuses    : ['complete', 'complete', 'not_run']
   RUN-LEVEL status     : partial
   estimated_cost_usd   : None     <-- withheld from a run every receipt supports
```

After: A is `partial`; B is `complete` with `estimated_cost_usd: 60.0`.

## This is a port, not a new rule

Both other summarisers over these same receipts already read the receipts' own states.

- `core/cost_receipts.py::_summary_status` drops `not_run` into a `contributing` list first.
- `scripts/cost-receipt.mjs` (shipped in #270) applies the same rule and explains it in a comment describing this exact failure — *"telling the reader that a run of 17 graded tasks and 1784 calls had never happened"* — and states it *"mirrors `_summary_status` in batch-runner/core/cost_receipts.py"*.

The JS comment on the neighbouring branch ends *"Mirrors `summarize_cost_receipts` in batch-runner/core/cost_projection.py"*. On the `rowsWithoutAReceipt` branch it genuinely does; on the status derivation it did not. That half-truth is what made the pair look settled.

## Measured, not asserted

Every real per-task receipt set under `data/grades`, folded four ways:

| summariser | verdict on the 21 published sets |
|---|---|
| `scripts/cost-receipt.mjs` (dashboard) | **`partial` × 21** |
| `core/cost_receipts.py::summarise_receipts` | **`partial` × 21** |
| `core/cost_projection.py` **before** | `not_run` × 20 · `partial` × 1 |
| `core/cost_projection.py` **after** | **`partial` × 21** ✅ |

`scripts/aggregate-grades.mjs:581` builds `costSummary` from those same per-task receipts, so **the dashboard has been showing `partial` for all 21 published shards while the Python report called 20 of them `not_run`.** One set of receipts, three answers. Now one.

**Blast radius on committed files: zero.** `summarize_cost_receipts` is reached only from `step3_format_results.py:270` and `step6_report.py:472`, both inference-side, whose output lands in `results/` on the runner rather than in this repository. The committed grade payloads carry a summary written by `summarise_receipts`, which has always applied the rule this ports. Measured with a repo-wide scan for run-level summaries: **0 found.** Both measurements are kept as tests, so a future edit fails with the path in hand.

## The change

```python
contributing = [r for r in receipts if r["status"] != "not_run"]
if not contributing:                                          status = "not_run"
elif all(r["status"] == "complete" for r in contributing):    status = "complete"
elif all(r["status"] == "unavailable" for r in contributing): status = "unavailable"
else:                                                         status = "partial"
complete_run = status == "complete"
```

`partial` is a label, not a number: an all-unpriced run still publishes `estimated_cost_usd: null`, `avg_cost_usd: null`, `max_cost_usd: null` and `measured_tasks: 0`. Nothing is invented to go with the new status. #315's demotion for rows carrying no receipt at all is byte-for-byte untouched and under test here — a *missing* receipt is a hole in the record, unlike a task that never ran, and the two must not be conflated.

## Mutation battery

12 mutations of the block, each run against the new file and against the 116 pre-existing tests in the four directly-affected files:

| # | mutation | new file | pre-existing 116 |
|---|---|---|---|
| M1 | the original bug, restored verbatim | **caught** | **not caught** |
| M2 | `not_run` filter dropped | caught | caught |
| M3 | emptiness tested on `receipts` — an all-idle run reads `complete` | caught | caught |
| M4 | `unavailable` branch removed | caught | caught |
| M5 | `all` → `any` for complete | caught | caught |
| M6 | `unavailable` skipped too | caught | caught |
| M7 | filter inverted — only idle receipts contribute | caught | caught |
| M8 | over-blocking: an idle-only run claims `partial` | caught | caught |
| M9 | over-blocking: no run is ever a total | caught | caught |
| M10 | every run is a total | caught | caught |
| M11 | `complete_run` back on the old count — label and total part company | **caught** | **not caught** |
| M12 | always `partial` | caught | caught |

**12/12 caught, 0 survivors.** The two the pre-existing suite could not see are exactly the two defects fixed here — which is also the cleanest statement of why no existing test had to change. M8 and M9 are the over-blocking negative controls.

## Verification

| check | result |
|---|---|
| full `pytest` | **6197 passed · 9 skipped · 45 deselected · 0 failed** (12m28s). Collected 6251 = prior 6199 + 52 |
| new file alone | 52 passed (9.7s) |
| the four affected files | 168 passed — the 116 pre-existing tests unchanged |
| mutation | 12/12 caught, 0 survivors |
| `mypy core/ --ignore-missing-imports` | 170 errors in 32 files — **baseline, unchanged** |
| `node --test 'scripts/__tests__/*.test.mjs'` | 193 · 192 pass · 1 skip — **baseline**; JavaScript is byte-for-byte unchanged |
| `tsc --noEmit` | clean |
| `npm run build` | clean |
| browser test | **NOT RUN** — `@playwright/test` is declared in `package.json` but absent from this machine. Recorded as not runnable, not as a pass |

Skip counts move by a few between full runs on this machine depending on Docker/seccomp availability; the collected total is what matches exactly.

## Grader source fingerprints

`core/cost_projection.py` is inside `compute_grader_source_hash`, so all four move. Computed with the real function; the before column was recomputed from a separate checkout of `3302399` rather than copied.

| config | before (`3302399`) | after |
|---|---|---|
| `default_v2` | `461dc8dea13fb3aa` | **`41fe8eb6d8dcec5b`** |
| `gold_smoke_audio_v2_sol_max` | `31752df82f934f52` | **`1ad880ccc1f4546c`** |
| `gold_ceiling_185_v2_sol_max` | `a85c48d8765b2b95` | **`f59eaa05bce2128a`** |
| `default_v2_sol_max` | `7ab61eaa96697e7f` | **`a76802c99768e23b`** |

All four before-values match §13.38's after-values exactly, so the stacked chain is unbroken. Test files are not in the hash.

## Do not merge

#292 · #294 · #295 · #297 · #299 · #301 · #312 · #313 · #314 · #315 · #317 · #318 · this one are **all held**. Nothing merges until every fingerprint is recomputed on current `main` with the real hash function and Session A's paid run has its evidence pinned. No model calls, no grading runs, no approval amounts, no Azure logins were involved in this change.


hyeonsangjeon:
**#336 으로 대체되었으므로 닫습니다.**

왜 개별 병합이 아니라 통합이었는지:

---

PR #409, opened 2026-09-03
fix(analysis): refuse a cohort that records no routing modality

hyeonsangjeon:
## What was wrong

`batch-runner/scripts/analyze_repeat_variation.py` proves §11 of the 315 preregistration — *this cohort has no `audio` items, so the audio flipping measured in 307/310–312 cannot be inside these intervals* — with one comparison:

```python
audio = [k for k, item in index.items()
         if item.get("routing_modality") == FORBIDDEN_MODALITY]
```

`routing_modality` is **not** in `schemas/grade.schema.json`'s required item fields. When it is absent, `.get()` returns `None`, `None != "audio"`, and the item is counted as *proven not audio*.

So the analysis gave the **same answer** for two different cohorts: one where every item was routed and none went to audio, and one where nobody recorded a route at all.

The prereg does not permit that. §11: *"이건 약속이 아니라 검사 가능한 사실이다 … 분석 스크립트는 이 코호트에 audio 항목이 0건임을 **확인하고**, 0이 아니면 멈춘다."* §12's closing line: *"0이 나오는 것과 0을 발표하는 것은 다르다."*

### The contrast that makes it a defect rather than a scoping choice

Three lines above, in the same loop over the same items:

```python
strange = sorted({str(item.get("verdict")) for item in index.values()
                  if item.get("verdict") not in VERDICT_VOCABULARY})
```

Delete `verdict` and the run **stops** — `str(None)` is `"None"`, outside the vocabulary. Delete `routing_modality` and the run **proceeds and reports an interval**. Same absence, same function, opposite treatment.

§12's reverse-check list includes "audio 항목 주입" but has no *unrecorded modality* mutation, which is why the gap was never exercised.

## How reachable it is

Measured over all 99 committed grade JSONs:

| | count |
|---|---|
| rubric items | 151,659 |
| `routing_modality` key absent | 42,717 |
| explicit `null` | 7,578 |
| **unrecorded total** | **50,295 (33%)** |
| `"audio"` | 365 |
| some other recorded value | 100,999 |
| **grade files where every item is unrecorded** | **11** |

Behaviour on the real published cohort before this change, `shape_problems` called directly:

| scenario | pre-fix result |
|---|---|
| A. untouched cohort (control) | `[]` clean |
| B. `routing_modality` key deleted, 4,299 items | **`[]` clean** |
| C. `routing_modality = None` everywhere | **`[]` clean** |
| D. one genuine `audio` item (control) | caught |
| E. `verdict` key deleted (control) | **caught** |

B and C are the defect; E is the same absence in the neighbouring check, caught.

## The change

`_records_modality(item)` — a route is a non-empty string — plus a **second, separately worded** refusal in `shape_problems`:

> `<run> records no routing modality for N rubric item(s), for example …. Section 11 is checked by reading that field, so an item that leaves it unrecorded leaves the claim that this cohort has no audio item unchecked rather than confirmed`

Kept apart from `contains N audio item(s)` on purpose. *Found the forbidden thing* and *could not look* are different answers, and a reader who gets one message for both learns nothing from either.

### Deliberately NOT a modality vocabulary check

The verdict check beside it pins its permitted set. This one must not, and that is the main design decision here. Any **recorded** route already answers §11's only question — that it is not `audio`. Pinning the permitted set would refuse a later cohort for naming a route this file has not heard of yet, which is a complaint about this file rather than about the cohort. A test locks that in: a cohort routed entirely to an unknown modality is accepted.

### Deliberately NOT changed

`:853` still does `str(item.get("routing_modality"))` when building the modality census. Once the guard refuses, `main()` raises at the gate before the census renders, so changing it would be unreachable code. Left alone to keep the diff to what the defect requires.

The preregistration itself is **not** edited. It is a document written before the analysis existed; amending it now to describe a mutation discovered afterwards would make it a worse record, not a better one.

## Published numbers do not move

All three payloads of the pinned 315 cohort record a modality for **every** item:

```
{'text': 1286, 'formatting': 69, 'visual': 77, 'mixed': 1}   = 1,433 items
zero absent, zero null
```

The new refusal has nothing to fire on there, so the interval in `data/grades/_validation/PR3_REPEAT_VARIATION.md` is arithmetically untouched — and `test_repeat_variation_report_quotes_its_run.py`, which re-runs the pinned command and compares the fenced block **byte for byte**, still passes. A new test asserts the zero-unrecorded property directly, so if it ever stops holding, the report stops being vouched for rather than quietly re-computed.

This is a latent defect in the sense of #150/#151: fixed for the payloads that come next, not to correct a published figure.

## Grader source hash: overlap 0

`compute_grader_source_hash` (`step8_grade.py:161`) reads 107 files + the grading config, all under `batch-runner/` — `step8_grade.py`, `core/**/*.py`, `schemas/grade.schema.json`, `requirements.txt`, `scripts/download_inference_from_hf.py`, and the prompt templates. **`batch-runner/scripts/**` is not among them**, and `batch-runner/tests/**` is not either.

Both changed files are outside it. The fingerprint cannot move, so **no re-grade is implied and this is safe to merge while shards are in flight.**

## Verification

| check | result |
|---|---|
| the 4 repeat-variation test files | **142 passed** |
| new tests | **8** |
| red-check: new tests vs. pre-fix source | **7 of 8 fail** |
| — the 8th | the vocabulary-freedom control, must pass **both** sides |
| backend pytest, full | **7,639 passed**, 10 skipped, 0 failed (17:12) |
| collected-test delta vs `HEAD` | exactly **+8** (7,647 vs 7,639) |
| repo-root `scripts/__tests__` | 183 passed |
| frontend `npm run test:aggregate` | 400 pass, 0 fail, 1 pre-existing skip |
| `tsc --noEmit` + production build | pass |
| `mypy` on the changed script | **Success: no issues found** |

The red-check was run without `git stash` — the file was copied aside, reverted with `git checkout HEAD --`, exercised, and copied back. Stash count re-verified at 3 afterwards.

## Cost

**0.** No model call, no re-grade, no dispatch. Reading, executing and testing only.


hyeonsangjeon:
### Merged-tree verification

`origin/main` moved to `928ae67` (#408) after this branch was cut, so it was merged forward (`1c07876`, merge not rebase) and the suite re-run on the merged tree:

| check | result |
|---|---|
| backend pytest, full, merged tree | **7,665 passed**, 7 skipped, 0 failed (18:06) |
| CI `pytest` | pass (16m57s) |
| changed-file overlap with #408 | 0 |

One note worth recording, because the first merged-tree run looked like a failure and was not.

That run reported `1 failed, 7664 passed`, on `test_mutation_d_resampling_items_is_marked_unregistered` — a pre-existing test, not one of the eight added here. The traceback:

```
scripts/analyze_repeat_variation.py:612: in item_bootstrap_rate
    hits += flags[rng.randrange(size)]
step8_grade.py:469: in _exit_on_signal
    sys.exit(128 + signum)
E   SystemExit: 143
[signal] received 15; unwinding so exit handlers can run
```

An external SIGTERM reached the pytest process. `step8_grade.py` installs a process-wide signal handler at import, so the signal surfaced as a `SystemExit` inside whichever test held the CPU at that instant; `test_mutation_d` runs a bootstrap resample loop and is the longest-running test in the suite, which is the whole of why it was the one hit. The run continued and everything after it passed.

Re-run detached in its own session: **7,665 passed, 0 failed** — 7664 + 1, the same total, with that test now passing. The file also passes 68/68 standalone. Nothing here is attributable to this change, and `step8_grade.py` is untouched (it is inside the grader source hash).

PR #447, opened 2026-09-07
diag(335): make the observation arm's format failure visible

hyeonsangjeon:
## What this is

Not a re-run of [334](../blob/main/tasks/rebuilding_grading_task/334-the-arm-that-broke-the-format.md). 334's 120 calls are a finished experiment; this places at most **8**.

334 reported that **51 of the observation arm's 60 replies** came back `format_error:unparseable_json`, and could not say why. That run stored no response bodies — `core.perception.audio` labels a malformed envelope and drops it — so the only surviving fact was the label. Re-reading 334 will not produce a body it never kept.

`scripts/replay_audio_format_failures.py` already answers everything the recorded *outcomes* can answer, at zero calls, and says so in its own docstring: *"The raw reply text was never stored… so nothing here re-parses JSON."* What is missing is the text.

This PR adds the diagnostic and its pre-registration. **It buys nothing.** The paid run happens after this merges, from the merge commit, as a separate dispatch.

## Files

| | |
|---|---|
| `tasks/rebuilding_grading_task/335-why-the-format-failed.md` | The pre-registration. Model, config, manifest, three claims, arms, the 8-request cap and the excerpt rules, all fixed before anything is bought. |
| `batch-runner/scripts/diagnose_audio_format_failure.py` | The collector. |
| `.github/workflows/audio-format-diagnostic.yml` | Free dry run → `grading` gate → paid job. |
| `batch-runner/tests/test_audio_format_failure_diagnostic.py` | 79 tests, no credentials, no calls. |
| `334-…md`, `CHANGELOG.md`, `.gitignore` | A correction record, the entry, and two `!negation`s. |

## What it collects, and what it will not

Default is **structural metadata only**: `finish_reason`, whether a refusal field was present and how long, the Python type of the content, which field the text was read from, length and sha256, first/last characters, fence presence + position + language tag, brace balance, the `JSONDecodeError`'s `msg`/`pos`/`lineno`/`colno`, core's own label, and which of the five contract keys were present and of what type.

Never collected: **the model's reasoning blocks** (the collector reads `message.content` and `message.audio.transcript` and there is no third place; a reasoning block leaves one boolean), credentials, personal documents, or any full reply.

Response text can be **excerpted at 240 characters maximum** into the committed report — masked *before* it is truncated, because the other order leaves half of a 60-character key in the file. Excerpts never reach stdout or a CI log. The input is this repository's own synthetic eSpeak clips.

## The cap is a reproduction budget

Eight requests including retries, counted in `DiagnosticClient.create` **before** each call, and a failed request still counts — a request that timed out may well have run. If the failure does not reproduce in the fixed sample, the run reports **not reproduced** and stops. There is no ninth request to go looking with.

The plan: 3 reproduce (observation arm) → 2 contrast (production arm, same claims) → 2 compatibility (`json_object`, `json_schema`) → 1 reserved for a single transport retry.

## What the sample cannot say

The three claims were chosen **because they failed in 334** — all three repeats `unanswered`/`read_failure`. A count over them describes the choice, not the arm. That sentence is stored next to every aggregate the run writes, and `test_the_fixed_sample_is_what_the_rule_selects` re-derives the sample from 334's raw data rather than trusting the constant.

## Structured-output compatibility: the documentation does not settle it

| Source | On `input_audio` + `response_format` |
|---|---|
| Azure Foundry Chat REST v1 | Both documented in the same request schema. **No statement either way** about combining them. |
| OpenAI Audio guide | Never mentions `response_format`, structured outputs, or JSON mode. |
| OpenAI Structured Outputs guide | Lists text models. Audio appears as neither supported nor unsupported. |

Azure's *images and audio* reference also has a `response_format` — that is `/audio/transcriptions`' `audioResponseFormat` (`json`/`text`/`srt`/`verbose_json`/`vtt`). Different endpoint, different parameter; not evidence here.

So probes 6 and 7 ask directly, and are reported **separately** from the reproduction result. **A rejection is the finding.** There is no parameter-free retry path in the module, and `test_there_is_no_parameter_free_retry_anywhere_in_the_module` keeps it absent — dropping the parameter and succeeding is not "supported", it is a different request. Passing probes 6/7 would mean the model *can be shaped*, not that it hears well; the latter is what 334 measured and this adds nothing to it.

## Correction to 334 §3

> **실패한 응답은 잘린 게 아니라 짧다.** 토큰이 모자라 JSON이 중간에 끊긴 모양이면 실패 쪽이 더 길어야 하는데, 반대다.

Output-token count is a **proxy** for truncation, not evidence of it. The field that settles it is `finish_reason`, and that run did not store it — so the correct answer is **"unknown"**, not "not truncated". 334's text is preserved; the correction is recorded beside it and in §7 here. `_truncation_verdict` reads `finish_reason` alone and returns `unsettled` rather than filling the gap with token counts, held by `test_truncation_is_read_off_finish_reason_and_never_off_token_counts`.

**No number moves.** 51/60, 12/20 vs 5/7, p = 0.0654, the 27-failure streak, 185 runs / 31 claims — all unchanged.

## What this does not touch

- **The other agent's grading run.** Separate workflow; its config, outputs and cost records are untouched. (It shares the accuracy probe's concurrency group *deliberately* — a diagnostic running alongside a 120-call probe would collect that probe's throttling and record it as a transport failure of its own.)
- **`core/`.** Not one line. Collection happens by wrapping the client core is already handed, so `compute_grader_source_hash` still returns `7506ce5008bd…`, byte-identical to what 333 §2 pins. Verified at this commit with the repo's own `probe.grader_source_hash`.
- **Grading configuration.** `gold_audio_repeat_v2_sol_max.yaml` is read, never written.
- **Stability figures and response-rate baselines.** Untouched.

## Cost

`gpt-audio-1.5` has no registered price in this repository. The report writes `pricing_complete: false` and `estimated_cost_usd: null`, and records input/output/cached/audio tokens in full. **It does not write `$0`** — filing a run that placed paid calls as zero is worse than filing nothing.

## Verification

- `pytest tests/test_audio_format_failure_diagnostic.py -q` → **79 passed**
- Full backend suite → **8,527 passed**, 12 skipped
- YAML parses; all 12 shell `run` blocks pass `bash -n`; both embedded Python heredocs pass `ast.parse` — and the tests assert the summary steps only read fields the report actually has, which is how the paid job's `if: always()` summary stops being the place a typo is discovered.


hyeonsangjeon:
## The `pytest` failure is on `main`, not in this branch

`pytest` here is red with **1 failure + 4 errors**. All five are about `data/grades/**` and the cost record that counts it. None of them name a file this PR adds.

### Reproduced with none of this branch's files present

```
git archive origin/main | tar -x -C /tmp/mainproof   # 602cea4, 1218 files, zero of mine
cd /tmp/mainproof/batch-runner && pytest -q \
  tests/test_both_cost_halves_recompute_from_their_ledgers.py \
  tests/test_the_grading_ledger_no_longer_dies_with_the_runner.py
```

```
ERROR ...::test_grading_total_recomputes_to_the_receipt
ERROR ...::test_both_halves_are_about_the_same_one_task
ERROR ...::test_every_multiple_the_config_states_recomputes
ERROR ...::test_neither_half_stands_a_zero_in_for_an_unpriced_call
FAILED ...::test_the_recorded_counts_recompute_from_the_ledgers
```

Same five, byte for byte, on a tree that has never seen this branch.

### What actually moved

`94bb8fa` + `602cea4` published a grade run for `exp026c_cost_receipt_smoke`, and two things followed:

**1. Two grade JSONs where the fixture expects one.** Same experiment, same rubric, same inference, same `_diagnostic/17e2607b…/` directory — differing only in `cfg_` and `src_`:

```
…__cfg_5b77131b1e1fc221__…__src_f4931215d1ec316b__v2.2.json
…__cfg_c4348e8fa153ce8d__…__src_7ca55f907056df2d__v2.2.json
```

`test_both_cost_halves_recompute_from_their_ledgers.py:109` resolves one and gets two, so its four dependents error at setup rather than fail.

**2. The committed ledger count is one behind.** `model_price_table.json` → `models_deliberately_not_priced['azure:gpt-5.6-sol']['measured_from_committed_ledgers']` records `ledgers: 32, rows: 68610`; the tree now holds **33** ledgers and **68,722** rows.

### Why nothing caught it before this PR

Both commits were pushed by the grading workflow, so they did not trigger `Backend Tests` — `gh run list --branch main` has no run for either SHA, and the last green one is `89e242d`, the commit before them. This PR is simply the first thing to run the suite against that tree.

### Not fixed here, deliberately

The ledger count lives in the other agent's cost record, and the test that catches it asks for a judgement I am not the one to make:

> Write 33 into `models_deliberately_not_priced['azure:gpt-5.6-sol']['measured_from_committed_ledgers']['ledgers']` **once you have checked that the prose around it still reads true.**

The `what_changed` prose beside that number narrates how many runs of this corpus exist and what the recount settled — a third one changes that story, and editing the integer without the prose would leave a record that counts correctly and reads falsely. The duplicate grade JSON is the same: whether it is a collision to resolve or two legitimate configurations is a question about that run, not about this diagnostic.

So this branch stays as it is. `advance-check` is green, and the seven files here are unchanged from the run that went **8,527 passed** locally against `89e242d`.

PR #473, opened 2026-09-09
feat(audio): run 344's verdict where the run happens

hyeonsangjeon:
## 무엇을 하나요 (What this does)

`verify_format_pilot_run.py` shipped in #470 and **nothing calls it.** A checker
that sits in the repository and is invoked by no job is a §0 precondition
satisfied on paper. This wires it into the two places the run actually happens.

344 §0 ticks boxes on two checks that only exist in CI — the free rehearsal, and
the paid pilot's §6 reading. Neither can be done on this box: there is no eSpeak
here, so the clips do not exist locally.

## The two steps

| job | reads | writes | best reachable verdict |
|---|---|---|---|
| `dry-run` | `audio-accuracy-dry-run.json` | `format-pilot-v3-rehearsal-verdict.{json,txt}` | `rehearsal_ok` |
| `measure` | `audio-accuracy-measured.json` | `format-pilot-v3-verdict.{json,txt}` | `format_held` |

Both are gated on `inputs.corpus == 'speech-format-pilot-v3'`, so every other
corpus is untouched.

**The filenames are the trap.** The two jobs write reports under different names.
Pointing the rehearsal step at `audio-accuracy-measured.json` would make it skip
on *every* rehearsal — and skip **silently**, because the missing-file branch
exits 0 on purpose (a run that stopped before writing a report should not be
reported as a failed verdict). Green tick, no verdict, nobody the wiser. A test
asserts each step names its own file and not the other one.

## Why the exit code is carried

Both steps run under `always()` and end `exit "$rc"`, after the summary is
written.

A rehearsal that came back `inconclusive` — wrong pin, short call count, no
ledger — under a green tick would clear a paid dispatch **using the very finding
that should have stopped it.** The summary goes first so the reason is on the
page when the step is red.

## Why the verdicts ride with the ledger

`337` is the reason. Its per-call record existed only in a CI artifact, the
artifact expired, and `340` had to be written to say the numbers could no longer
be produced. A verdict in a step log has exactly that lifetime. So each verdict
file is added to the upload that already carries the ledger rows it was computed
from — the two cannot be separated later.

## The four break-tests

Each was run against a deliberately broken workflow and each went red; the file
was then restored byte-identical (`diff -q`).

1. rehearsal step pointed at the paid filename → caught
2. exit code swallowed → caught
3. verdict files dropped from the upload → caught
4. `always()` removed → caught

## Also: a docstring that named the wrong file

`verify_format_pilot_run.py:9` said it reads "the report and the ledger export
beside it." It does not. It reads the report and the pre-registration; the
ledger *reference* comes from inside the report's own cost block, which is what
condition 7 reads — which is why a report stays checkable long after the
`.sqlite3` beside it is gone. `grep` confirms only two read sites exist in the
file. For a tool whose entire case rests on naming exactly what it opens, that
sentence had to be right.

Found while reading my own merged code in #470; deferred to here rather than
restart that PR's CI.

## Verification

- new tests: **4 passed**
- the two touched files: **254 passed, 1 skipped** in 24.31s
- full backend suite: **9255 passed, 12 skipped, 45 deselected** in 19:34, exit 0
- YAML parses; 4 jobs; no duplicate step names
- `git diff --name-only origin/main | grep batch-runner/core/` → **none** (lane separation with A's Codex work)

## Not fixed here, on purpose

`main` has no CI trigger. Two individually-green PRs can therefore leave `main`
red — as happened at `446acfd` for ~70 minutes on 2026-09-08, which is what the
CHANGELOG `### Fixed` entry explains. Fixing it means editing the workflow
triggers, which is A's lane this week. Recorded, not touched.

## Cost

**Nothing is dispatched and nothing is bought by this PR.** It makes the *free*
rehearsal readable, which is the next §0 box. Five §0 boxes remain ⬜ before any
paid call is possible.

### DIFF
--- a/.github/workflows/audio-accuracy-probe.yml
+++ b/.github/workflows/audio-accuracy-probe.yml
@@ -352,6 +352,54 @@ jobs:
               print(f"- pairs available for the arm comparison: {comparison['pairs']}")
           PY
 
+      # 344 §6's verdict, on the rehearsal. The same checker the paid job
+      # runs, with the same seven conditions, so "the plan is ten requests and
+      # the run makes ten" gets settled here for nothing rather than
+      # discovered on the dispatch that was bought to answer something else.
+      #
+      # What a rehearsal cannot answer it says it cannot answer. The replies
+      # come from a stub, so the two arm conditions come back `unanswerable`
+      # and the best verdict reachable here is `rehearsal_ok`. A free run
+      # reporting `format_held` would be the experiment replaced by its own
+      # scaffolding, and the checker has no flag that permits it.
+      #
+      # The exit code is carried rather than swallowed. This step ticks a §0
+      # precondition, and a rehearsal that came back `inconclusive` -- wrong
+      # pin, short call count, no ledger -- under a green tick would clear a
+      # paid dispatch using the very finding that should have stopped it. It
+      # is carried *after* the summary is written, so the reason is on the
+      # page even when the step is red.
+      - name: Check the rehearsal against 344 §6
+        if: ${{ always() && inputs.corpus == 'speech-format-pilot-v3' }}
+        run: |
+          set -uo pipefail
+          cd batch-runner
+          if [ ! -f "$GITHUB_WORKSPACE/audio-accuracy-dry-run.json" ]; then
+            echo "::warning::no report, so there is no rehearsal to check"
+            exit 0
+          fi
+          python scripts/verify_format_pilot_run.py \
+            "$GITHUB_WORKSPACE/audio-accuracy-dry-run.json" \
+            --out "$GITHUB_WORKSPACE/format-pilot-v3-rehearsal-verdict.json" \
+            | tee "$GITHUB_WORKSPACE/format-pilot-v3-rehearsal-verdict.txt"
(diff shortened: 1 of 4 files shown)


hyeonsangjeon:
Correcting one claim in this PR's description and in the CHANGELOG entry it landed.

I wrote that `main` has no CI trigger, so nothing reported the red state. **That is wrong.** `backend-tests.yml` carries `push: branches: [main]`, and commit `446acfd` started run [`34332874303`](https://github.com/hyeonsangjeon/gdpval-realworks/actions/runs/34332874303), which finished `failure` at 09:28 naming the exact assertion (`assert 'ee1ca0b0…' == '7e745a18…'`, 1 failed / 9195 passed). CI said so inside twenty-two minutes; nobody read it. I noticed because merging this PR fired two runs on `main`, which cannot happen if there is no trigger.

The two real mechanisms — a paths filter that excludes `tasks/**`, and `cancel-in-progress` leaving `main` commits with no verdict — are measured and recorded in #474, along with the free `workflow_dispatch` remedy. Nothing in this PR's code is affected; the error was in the prose.

PR #476, opened 2026-09-09
docs(344): tick the two boxes the rehearsal filled, and only those

hyeonsangjeon:
## What this is

`344`'s free rehearsal ran. This ticks the two §0 boxes it filled — **and no others.**

Run **`34351518357`** on `6e867a66`, dispatched with `dry_run: true`,
`paid_approval: false`, `repeats: 1`, `prompt_arm: production`,
`corpus: speech-format-pilot-v3`. All three paid jobs report `skipped`:
`approve-paid` is gated on `dry_run == false && paid_approval == true`, and a
skipped dependency skips `measure`. **No path to a billed call existed, and
nothing has been bought.**

## What the verdict says

`VERDICT rehearsal_ok — 5 passed, 0 failed, 2 unanswerable`

| # | condition | result | value |
|---|---|---|---|
| 1 | candidate header is the registered one | pass | `dd0380fa…` in report and prereg |
| 2 | grader is the pinned one | pass | `7e745a18…` in both |
| 3 | the ten planned requests went out | pass | 10 requests / 10 calls / no retry / no early stop |
| 4 | both arms ran the same five items | pass | §4's five, in §4's order |
| 5 | production arm readable ≥4/5 | **unanswerable** | replies came from the stub |
| 6 | candidate arm readable 5/5 | **unanswerable** | same — and this is the question being bought |
| 7 | the run wrote a cost ledger | pass | 10 rows, receipt `partial`, amount `null` (`price_missing`) |

**5 and 6 coming back unanswerable is the correct outcome.** A rehearsal that
passed those two would be reading canned stub text as evidence about
`gpt-audio-1.5`, and would remove the reason to buy anything. That is why the
verdict is `rehearsal_ok` and not `format_held`.

Amount `null` is also correct: `gpt-audio-1.5` is unpriced, the ledger says
`price_missing`, and that is **not `$0`**.

## What is still open

Three boxes, in this order:

- ⬜ re-measure §2's four fingerprints **at the dispatch SHA**
- ⬜ confirm A is not mid-`core/` change
- ⬜ pin that SHA behind an immutable tag (§12)

## Also in here

- §0's earlier boxes are attributed to the PRs that actually filled them —
  #468 (`446acfd`) for the header and corpus registration, #470 (`82db5fc`)
  for the response-shape record, #470 + #473 (`ce847ff`) for the verifier and
  its broken-input cases — rather than left blank.
- **First look at §9's response-shape record on a real artifact**, under
  `wire.response_shapes` in the report. It carries what §9 promised:
  `finish_reason`, brace counts with a balance flag and first-brace offset,
  the JSON error type with its **character position and characters
  remaining**, the five contract keys as present/missing with types, char
  count, first and last non-whitespace character, and a masked excerpt (162
  chars, under the 240 ceiling). Reasoning is a **boolean**
  (`reasoning_present`); no credential, no request id, no reply body beyond
  the excerpt. This is exactly what 337 spent ten calls without recording.
- **Two corrections to §0's paragraph on paths-filtered commits, both from
  things that happened while this PR was open.**
  - #475 (`0aae1c3`) put `tasks/**` into both triggers, so the box-7 re-pin
    commit now starts a run on its own. That paragraph becomes a record of
    how it used to be, with #469 still the evidence. The explicit dispatch
    stays anyway — §12 wants a run held to **one commit**, not to a branch
    that moves.
  - The dispatch meant to give `main`'s merged commit a verdict,
    `34351414118`, came back **cancelled**: `concurrency:
    backend-tests-${{ github.ref }}` with `cancel-in-progress`, and #475's
    own push run took the group. So the pre-dispatch check is not "a run was
    started" but **"that run finished `success`"**. #475 did not touch the
    concurrency block.

## Checks

Rebased onto `0aae1c3`, so #475's widened filter applies and **this PR now
starts a `pytest` run by itself** — which is the first live confirmation of
that change on a `tasks/**` + `CHANGELOG.md`-only PR. Locally: 307 passed,
3 skipped across the suites that read either file.


hyeonsangjeon:
Dispatched `backend-tests.yml` at `f28bf81e` — run [`34352411123`](https://github.com/hyeonsangjeon/gdpval-realworks/actions/runs/34352411123), since the paths filter gives this PR no checks of its own.

The rehearsal artifacts are on run [`34351518357`](https://github.com/hyeonsangjeon/gdpval-realworks/actions/runs/34351518357): `format-pilot-v3-rehearsal-verdict.{json,txt}` and the ten-row ledger, beside `audio-accuracy-dry-run.json`. The verdict text is quoted in full in the PR body table.

PR #478, opened 2026-09-09
docs(344): re-measure the four pins at the dispatch SHA, and say why the ninth box stays open

hyeonsangjeon:
## What this is

The last preparation PR for `344`. It fills §0's **seventh and eighth** boxes
and explains why the **ninth cannot be filled from inside this document**.

**Nothing has been bought.**

## Seventh box — the four pins, re-measured at `cff4d7c1`

| line | document | measured now | how |
|---|---|---|---|
| grader | `7e745a18…` | `7e745a18…` | `grader_pin_stated_in` ↔ `grader_source_hash()` |
| price table | `b01b384c…` | `b01b384c…` | `price_table_pin_stated_in` ↔ `sha256(PRICE_TABLE_PATH)` |
| manifest | `97755288…` | `97755288…` | `manifest_pin_stated_in` ↔ manifest bytes |
| candidate header | `dd0380fa…` | `dd0380fa…` | `candidate_pin_stated_in` ↔ `sha256(SPEECH_OBSERVATION_HEADER_V3)`, 1,931 chars |

**All four match, so no value in §2's table changes.** What was missing was
the measurement, not the numbers — §0's box asks for the act, and `#469`
patched one row as a red-`main` repair, which is not the same thing.

Both sides use the repository's own accessors. Hashing the files by hand
would have proved the two agree with *my* arithmetic rather than with the
code's, and the document-side value is exactly what the measurer will read.

This commit cannot move any of the four: it touches this document and
`CHANGELOG.md`, the grader fingerprint covers neither, and the manifest is a
file this work never opens. That is not left as an assurance — the measurer
re-checks the grader and header pins against the document at run time and
stops before calling the model if either has moved. Conditions 1 and 2 of the
rehearsal are that check running.

## Eighth box — `core/` is not moving

- **Zero open PRs** at `cff4d7c1`.
- `main`'s last `core/` change is `7a8c94d` (#467). **Ten commits** have
  landed since and none touches `core/` — #477 included, which is
  documentation only.
- The live grader hash is still the `7e745a18…` that #467 moved it to.

## Ninth box — left unticked, deliberately

The tag goes on the commit that carries this document. Ticking the box here
would seal a document claiming a finished step **inside the tag that step
creates** — sequence reversed, which is post-hoc justification wearing a
pre-registration's clothes.

So §12 now moves that box's evidence out of markdown and into `git`:

```
git rev-parse audio-344-dispatch^{commit}      # what the tag points at
gh run view <run> --json headSha               # what the run actually ran
```

Equal → bought. Different → **nothing is bought.** Both values go at the top
of §13. And a failed run does **not** get the tag moved to a new commit —
that is §7's forbidden re-dispatch under another name.

## Also

§12's opening no longer asserts A is mid-`core/`-change. The tag does not
guard against PRs that are open now; it guards against merges that land
**while a run sits in the queue**, which nobody can see in advance.

## Checks

Locally: 307 passed, 3 skipped across the suites that read either file. With
#475 merged, `tasks/**` starts a run on its own, so this PR has real checks.


hyeonsangjeon:
## Amended before dispatch: the tag cannot be the dispatch ref

While verifying the destination for the paid run, the plan in §12 turned out
not to execute. Recording it here because it changes this PR's content.

`approve-paid` deploys to the `grading` environment. That environment's
deployment policy allows **exactly one ref — the `main` branch**
(`custom_branch_policies: true`, one entry). A dispatch from
`audio-344-dispatch` would clear the free `dry-run` job and then die at the
gate.

**This is on record, not deduced.** Run `34038055900` (2026-09-06) dispatched
this same workflow from tag `speech-diagnostic-run-1` at `5418048`:

| job | result |
|---|---|
| Dry run (no model call, no cost) | success |
| Approve paid audio accuracy probe | **failure** |
| Measure (paid) | skipped |

> `Tag "speech-diagnostic-run-1" is not allowed to deploy to grading due to environment protection rules.`

The old §12 would have spent the one authorised dispatch on a gate error.

**The allow-list is not being widened.** Adding the tag changes what the paid
gate accepts — a protection-surface change to report with its scope, not to
make in passing. Exact action, recorded in the document for whoever decides:
settings → Environments → `grading` → deployment branch/tag policy → add
`audio-344-dispatch` as a **tag**. This work proceeds without it.

**The replacement checks more.** Tag still cut at the merge commit, as an
immovable name rather than a ref. Dispatch on `--ref main` while `main` still
points there → GitHub freezes `head_sha` at creation → read it back and
compare to the tag → **approve the paid gate only if they match.** `measure`
is behind `approve-paid`, so the comparison happens before any spending. The
old order took the commit on faith and checked afterwards.

Amended because the step does not run, not because a result came back.
Nothing has been measured and nothing has been bought.

Head is now `141ddce`; 315 passed / 2 skipped locally.

PR #485, opened 2026-09-10
feat(agentic-v2): stage A — the one paid question, built and refusable

hyeonsangjeon:
Two commits. The first works out what stage A's one paid question is allowed to
cost and builds the probe that asks it; the second builds the run that actually
asks, and the only place it can run.

Nothing has been dispatched. No model has been asked.

## Why this is a workflow and not a command

`AzureAIClientFactory` calls `_reject_static_azure_credential_env` before it
builds anything and takes its identity from a federated OIDC session. Measured
on this box: `AzureAIClientFactory()` raises `ValueError: AZURE_AI_ROUTE_PROFILE
is required`, and `env | grep -ci azure` returns 0. So a client cannot be built
from a checkout at all, and the only place the paid run works is a job holding
`id-token: write` — which runs what the repository contains. That is why
`scripts/run_agentic_stage_a_probe.py` is committed rather than kept local, and
why the workflow file is part of the guard rather than a convenience around it.

It also makes `--dry-run` free by construction rather than by intention: with no
Azure route configured a client cannot be built, so a dry run that exits cleanly
is a dry run that never tried. One of the 33 tests demonstrates exactly that
rather than asserting it.

## What the gate decided

$0.59 at most — $0.47 running, $0.00 marking, 8 model calls maximum — against
$1.00 approved for stage A. Stage one's own `approved_maximum_usd` is untouched
and still `null`; this is a separate, much smaller purchase and `finalize` is not
among the offered tools, so nothing here can reach the grader.

## What has to be true for the run to count

Three things together: a model was reached, at least two turns happened, and a
later call went out holding an earlier tool's answer. Any one alone is true of a
run that proved nothing. A model that was reached and declined exits 1 — and its
record is uploaded anyway, because that is a finding and stage A collects
findings rather than laundering them into successes.

## What the run leaves behind

One row per model call: turn, deployment asked for, model that answered, token
counts, how much earlier conversation the call carried, and either a price or an
explicit `price_missing`. Not the benchmark wording, not a request or reply body,
not a credential. Tested, not just intended.

The task's wording is read from the pinned dataset at run time and hashed against
the catalogue, which holds the hash and never the text. A revision that resolved
to different text is caught before the first call rather than discovered by
paying for it. The resolved route is checked against all three things the plan
fixes — account, project and route profile — after the client exists and before
anything is asked, and all three mismatches are reported at once so a
misconfigured dispatch is one fix rather than three runs.

The workflow cannot spend by accident: `workflow_dispatch` only, `mode` defaults
to `dry-run`, the paid job needs both the mode and a passing free job, only it
holds a token, and runs are never cancelled part-way — a killed conversation has
been charged for turns it would leave no record of. The deployment is read from
the plan at run time rather than restated, so the route that gets validated
cannot drift from the model that gets asked.

## Shared files

`.gitignore` is shared with A's Codex work. The change here is one allow-list
line plus its comment, under the existing `batch-runner/scripts/*` deny rule —
without it the runner script is invisible to `git add -A` and the workflow would
call a file the checkout does not have. Both new files are in
`test_the_new_files_are_in_the_repository`, the test that exists to catch that.
CHANGELOG.md is deliberately untouched to keep this clear of A's branch.

## Tests

`test_run_agentic_stage_a_probe.py` — 33, nine of them on the workflow file
itself. Full backend suite green on the pre-rebase tree (9377 passed, 15 skipped)
and re-running on this one.

## Still not done

The dispatch has not been made and no model has been asked. Everything that
decides whether it may be is now built, tested and refusable.


hyeonsangjeon:
Third commit: the 59 cents checked against the raw data rather than against the
check that printed them.

Worked out again by hand from the sources — `1.25`/`5.00` per million from
`model_price_table.json`, `3.0` characters a token and `7307` characters of
instruction and the `1.25` multiplier from `advance_check_plan.yaml`, `1589`
characters of task from the catalogue, `65536` bytes of tool result from the
dispatcher's own dataclass, 4 calls across 2 attempts from the plan:

```
model calls          8
input tokens         310,456   (base 2,966/call, 23,894 carried per prior turn)
output tokens        16,384
to run               $0.46999
with x1.25 safety    $0.5874875
```

The free check prints `310456 tokens sent, 16384 tokens received` and `$0.47` /
`$0.59`. Same to the last digit.

It caught one thing. The arithmetic only lands there with `7307` characters of
instruction; at the `5020` the plan's own prose still claimed it would be
`$0.46`. That sentence went stale when the previously-uncounted sections were
counted, and it took the cheapest stage-one row with it — `3.37` to run, where
the prose said `3.32`. The figures were never wrong, only the sentence
describing where they came from, and nothing reads those words, which is why
they could drift. The sentence now names the history (1068 → 5020 → 7307) and
no current number, since repeating the current one just sets up the third drift.

hyeonsangjeon:
Fourth commit, and the reason the CI wait was worth spending on verification
rather than on watching: **the route check would have refused the paid run, and
the fault was in the check.**

Resolved the route the workflow really configures instead of the one the tests
imagined:

```
AZURE_AI_ROUTE_PROFILE=project-ci
FOUNDRY_PROJECT_ENDPOINT=.../api/projects/gdpval-realworks

profile        project-ci
endpoint.kind  direct-v1
account        'hjeon-fdpo-foundry-eus2'
project        None
url            https://hjeon-fdpo-foundry-eus2.services.ai.azure.com/openai/v1/
```

`project-ci` derives an account-scoped `direct-v1` URL from the project endpoint
and sends inference over it, so the selected endpoint carries the account and no
project. Checking the selection's project alone refused the one configuration
the paid job uses — after the Azure sign-in, before anything was asked. No money
lost; a dispatch spent reporting a fault in the check.

The project a run is held to is the one the client was built from, so it is read
from those settings when the selection does not carry it, and the runner now
reads the environment once and hands the same settings to both the factory and
the check. Verified against the real resolution:

```
with settings   : ACCEPTED
without settings: ["...neither the route that was selected nor the endpoints it
                    was built from names a project, so there is nothing to check
                    it against"]
wrong project   : ["the route resolves to project 'not-ours', but the plan fixes
                    'gdpval-realworks'"]
```

Not a fallback that looks away — a wrong project endpoint is still refused, and
a project nothing names is refused too, with its own sentence. 37 tests in the
file now.

hyeonsangjeon:
**The paid job would have failed before asking anything, and it was not the route check this time.**

`AZURE_AI_REQUIRE_EXPECTED_IDENTITIES: '1'` makes `AzureAIRouteSettings.from_env`
demand every name in `REQUIRED_IDENTITY_ENV_BY_PROFILE` for the profile. `project-ci`
lists three. Both steps that reach Azure passed two.

Measured, not read — the step's exact environment handed to `from_env`:

```
--- exactly what the workflow set ---
ValueError: required Azure AI endpoint identities are missing: AZURE_AI_EXPECTED_DIRECT_ACCOUNT

--- with AZURE_AI_EXPECTED_DIRECT_ACCOUNT added ---
built. settings.project.project = gdpval-realworks
route: kind=direct-v1 account='hjeon-fdpo-foundry-eus2' project=None profile=project-ci
```

That would have landed after the federated sign-in and before the question: a
spent dispatch reporting a configuration fault. Every other paid workflow in the
repository passes all three; this one was the exception.

The fix is one line in each of the two steps. The test that comes with it is not
about those two lines — it reads the required names out of the code's own table
and holds every step that switches the demand on to it, so a profile that grows a
fourth requirement fails on the day it grows one rather than on the day somebody
dispatches. Run against the file as it stood, it names both steps and the missing
variable.

**The `Validate the Azure AI route` step now rehearses green**, with the same
environment the job builds:

```
[{"endpoint_kind":"direct-v1","profile":"project-ci",
  "runtime_fingerprint":"bdf9393c…","workload":"inference"}]
exit=0
```

Endpoint-free, as intended. The `$GITHUB_ENV` hand-off before it produces exactly
`AZURE_AI_WORKLOADS_JSON=["inference=gpt-5.4"]`, and the second measurement above
confirms the route check added in `9880dc3` accepts the route this configuration
really selects.

Nothing has been dispatched and no model has been asked.

PR #486, opened 2026-09-10
diag(codex): bisect the served request against what a Codex turn adds

hyeonsangjeon:
## What this is

Run `34442249527` sent `--valid-request` to the pinned deployment and got **`200`,
`response_status: completed`, model text back, 13 input / 5 output tokens** — same
host fingerprint that answers the Codex runtime's `401`, token minted by the
runtime's own `auth_command`, same path, same scope, same deployment name.

That is the first servable request this route has ever received, and it retires
five hypotheses at once: wrong resource, missing inference permission, wrong
deployment name, bad token minting, wrong route. **It also retires the case for
asking anyone for an RBAC role — this identity demonstrably may infer here.**

So the open question narrowed to one: *which property of a Codex request stops a
request that otherwise works?* This PR builds the instrument that asks it.

## `--closing-sweep`

Nine servable requests. Arm 1 is the request run `34442249527` proved servable,
unchanged. Arms 2–9 are that request plus **exactly one** thing a real Codex turn
adds, measured off the wire by `test_what_one_codex_turn_actually_sends.py`:

| arm | added |
|---|---|
| `accept_event_stream` | `Accept: text/event-stream` |
| `originator_header` | `originator` + the runtime `User-Agent` |
| `session_headers` | the six session/turn headers |
| `stream_true` | `stream: true` |
| `store_true` | `store: true` |
| `codex_instructions` | a 19 KB `instructions` |
| `codex_tools` | ten tool definitions, `tool_choice: auto` |
| `everything` | all of it at once |

**Isolated, not cumulative** — a cumulative sweep's first flip is attributable
only to "this arm or an earlier one". One request per arm, no retry, 512-token
ceiling, 90-second wall clock, concurrency 1, ceilings in the script so a
dispatch cannot widen them.

**If the baseline is not served on the day, the other eight are not sent.** They
would be measured against nothing and they are not free; that run records
`closing_sweep_inconclusive`. A `400` on an arm is **not** counted as the gate
closing — that is the route objecting to the request's shape, and reporting it as
a permission fact is how somebody ends up asking for a role over a malformed
field. Finding nothing is recorded as finding nothing, which points at the
transport or at a body property this sweep does not vary.

## Two corrections

**1. A null read as a finding when the instrument could not see.** Run
`34437632382`'s free header sweep returned `properties_that_closed_the_gate: []`,
and this changelog, the job summary and the task doc all took that to mean the
runtime's headers are not what refuses it. Wrong. Every arm's body named no
`model` — that is what made them free — so every arm died in the deployment
router with `400 Missed model deployment` **before reaching whatever refuses the
runtime**. The empty list is the instrument failing to see, not a measurement of
nothing. The three places that said otherwise now say this; the old record is
kept with the reason it was wrong beside it.

**2. The diagnostic could write a record that was neither checked nor kept.** The
leak check's path list and the artifact upload's path list are written separately
in the same workflow and nothing compared them. A probe whose `--out` was added
to one and not the other either published an unchecked record — on a public
repository — or spent money and threw the evidence away, and the run stayed green
either way. Two tests now hold the lists equal and require every `--out` any step
writes to appear in both. Mutation-verified.

## Shared files touched

Per the two-session split, declaring these up front:

- **`.github/workflows/codex-foundry-connection-diagnostic.yml`** — A-owned
  (Codex diagnostic workflow), but listed because it is a workflow. Additions
  only: one `send_closing_sweep` input, one step, one leak-check branch, one
  upload path, one summary section. No change to any existing step's behaviour,
  no change to the `if: github.ref == 'refs/heads/main'` guard or the
  `runs-on: ubuntu-22.04` pin.
- **`CHANGELOG.md`** — shared. This branch merged `origin/main` (`a423244`,
  carrying #482/#483/#484) and resolved the conflict by keeping both sides:
  B's Agentic V2 stage-A entries are intact; the only line removed is the
  sentence correction #1 above describes.

Nothing else shared is touched. `batch-run.yml`, `executor.py`, the price table
and the ledger are untouched.

## What this does not do

- `urllib`, not the Codex runtime — a property that closes the gate is the next
  place to look, not a proven cause.
- No tool ran, no file was written: **no task was solved**. The 220-task column
  is untouched.
- `price_usd: null`, `pricing: partial` — across nine paid requests. Unpriced is
  not free.
- The sweep has not been dispatched. This PR is the instrument.

## Tests

- `tests/test_codex_closing_sweep.py` — 31 new tests; every host in them is a
  loopback socket on the test machine.
- `tests/test_codex_foundry_connection_probe.py` — 101 → 106.
- Affected files after the merge: **185 passed**.
- Full backend suite: green (reported in a comment below with the run's own
  numbers).
- `mypy` clean for `scripts/diagnose_codex_foundry_connection.py`.


hyeonsangjeon:
Full backend suite after the `origin/main` merge, run locally against the pinned venv:

```
9406 passed, 12 skipped, 45 deselected, 2 warnings in 1408.60s (0:23:28)
```

exit 0. The two warnings are the usual `PytestUnknownMarkWarning` for `@pytest.mark.timeout` — `pytest-timeout` is not installed in this venv, and the marks are inert rather than failing.

CI on `6a201f5`: `pytest` pass (21m57s, run 34446403435), `advance-check` pass (3m25s, run 34446403470).

Merge resolution, for the record: `origin/main` at `a423244` carried #482/#483/#484. Four of the five conflicted files were byte-identical between main and this branch's base — `git rev-parse` on the blobs confirmed it — because #483 landed the same content this branch already had, so their conflicts were spurious and were resolved to this branch's version. Only `CHANGELOG.md` had real divergence; both sides were kept and the single line removed is the sentence the correction above replaces. B's Agentic V2 stage-A entries are intact.

Re-derived rather than restated: `tests/test_codex_closing_sweep.py` collects 31; `tests/test_codex_foundry_connection_probe.py` collects 106 against main's 101, and the five added test functions are absent from main's copy. The merged workflow parses to 19 steps, keeps `runs-on: ubuntu-22.04` and the `refs/heads/main` guard, and its leak-check and upload lists are equal at 7 records with every `--out` in both.

PR #487, opened 2026-09-10
fix(agentic-v2): stop offering tools the service will refuse

hyeonsangjeon:
The first paid stage A dispatch — run [34450289535](https://github.com/hyeonsangjeon/gdpval-realworks/actions/runs/34450289535), 2026-09-10 — was refused with `400` before the model was asked.

**Nothing was charged.** The record the probe wrote, uploaded as the run's artefact:

```
"reached_a_model": false,   "model_calls": [],   "turns_taken": 0,
"spent_usd": "0",           "resolved_model": null,
"stop_reason": "model_stopped_without_finishing",
"detail": "the model stopped without committing an answer:
           asking the model failed: BadRequestError"
```

A `400` refuses the request, so the model is not asked and there are no tokens to bill. `model_calls` is empty because none was made and `spent_usd` is `0` because it is measured from the calls, not assumed. The workflow reported the failure honestly: `Ask` captured the probe's exit code and `Report what happened` re-raised it.

## What was wrong

Every tool was offered with `strict: true`. That flag is not a setting for how carefully arguments are checked — it declares that the schema fits a narrow subset the service enforces on the model's behalf, and the service validates the declaration when the request arrives. Of the two tools stage A offers, `workspace_apply` is a `oneOf` at the root (one branch per operation, each carrying only its own fields) and `capabilities_query` has genuinely optional fields. Both are outside the subset. **Neither could ever have been sent.**

An overclaim does not degrade to a laxer check. It makes the call impossible, before inference — which is why no amount of local rehearsal found it. The dry run builds no request, and every unit test read the definitions back rather than asking whether a service would take them.

## Nothing is loosened

`strict` is now `false`. What the desk will *act* on is unchanged: `validate_tool_arguments` validates against the same schemas in full — `oneOf`, `pattern`, length bounds and all — and remains the only thing between a tool call and the workspace. The schemas were deliberately **not** narrowed to fit the subset; they are the real contract. What is dropped is a promise that could not be kept. The cost is a turn occasionally spent on arguments the model must be told were malformed, which is the right price for a call that otherwise cannot be made at all.

## The second defect: the failure could not be read

The record said only `BadRequestError`, so the cause had to be re-derived by hand from the payload while the service had already answered in the status and error code it sent back.

The note now carries both, through the same reduction `code_interpreter` already used for provider refusals — status and a code-shaped code, and nothing else read off the exception. That reduction moves to `core/provider_refusal.py` rather than existing in two copies, because redaction kept in two places is redaction that eventually differs in one of them. An endpoint, an account, a project or a deployment name cannot pass a code-shaped allow-list, so a note that says *what* was refused still cannot say *who* refused it.

## Guards, both offline and free

- **`test_tool_definitions_can_honour_what_they_claim`** reads the definitions the way the service does and fails any that claims `strict` while breaking the subset. It includes a test that it fails on the exact definition that was refused, because a guard that has never failed on the real thing proves nothing.
- **The voice may only send keys `Responses.create` actually accepts**, so a misspelled parameter cannot reach a paid dispatch to be found. The fakes accept anything; the real signature is free to check.

## v1 has the same defect and is NOT fixed here

The guard covers v1 too, and found the identical overclaim in `run_ffmpeg`. An `agentic_sandbox` run that offers it is refused with `400` today.

It is not fixed in this PR. v1's tool JSON is hashed into `V1_TOOLS_SHA256` in `test_agentic_v2_compatibility`, which freezes v1 as the baseline v2 is measured against — changing it is a decision about that baseline and about every v1 result already recorded under it, not a side effect of a v2 fix. So it is recorded in `KNOWN_OVERCLAIMS` and asserted to **still** be broken: the day somebody fixes it, the test fails and puts the frozen hash in front of them.

## Overlapping file

`core/code_interpreter.py` is shared. The change is that it imports the reduction from its new home and keeps `_provider_error_classification` and friends as aliases. Behaviour is unchanged and its 30 existing tests pass untouched.

## Checked

- 310 tests across every affected file, green.
- `mypy` on `core/provider_refusal.py`: clean; the changed modules add no new errors.
- Full backend suite running locally alongside CI.

### DIFF
--- a/batch-runner/core/agentic_v2_contract.py
+++ b/batch-runner/core/agentic_v2_contract.py
@@ -430,6 +430,33 @@ class AgenticV2Lifecycle:
 
 
 def responses_tool_definitions() -> list[dict]:
+    """The tools as the Responses API is asked to present them.
+
+    ``strict`` is false, and deliberately. It is not a description of how
+    carefully the arguments are checked — it is a promise that the schema
+    fits the narrow subset the service will enforce for the model, and none
+    of these schemas do. ``workspace_apply``, ``exec_run`` and ``browser_run``
+    are a ``oneOf`` at the root, so each operation carries exactly the fields
+    that operation needs and no others; ``capabilities_query`` has genuinely
+    optional fields. Both shapes are outside the strict subset, which wants a
+    single object with every property required.
+
+    Claiming ``strict`` anyway does not degrade to a laxer check. The service
+    validates the schema against the subset when the request arrives and
+    refuses the whole call with ``400`` before the model is ever asked, which
+    is what happened to the first paid stage A dispatch on 2026-09-10: nothing
+    ran, nothing was charged, and the run recorded a model that never spoke.
+
+    Nothing is loosened by dropping it. ``strict`` only ever constrained what
+    the model was free to emit; what the desk is willing to *act* on is
+    decided by :func:`validate_tool_arguments`, which validates against these
+    same schemas in full — ``oneOf``, ``pattern``, length bounds and all — and
+    raises on anything that does not fit. That check is unchanged and remains
+    the only thing standing between a tool call and the workspace. The cost of
+    the change is a turn occasionally spent on arguments the model has to be
+    told were malformed, and that is the correct thing to pay for a call that
+    otherwise cannot be made at all.
+    """
     descriptions = {
         "capabilities_query": "Discover actual task-environment capabilities and remaining budgets.",
         "workspace_apply": "Perform one bounded operation in the task-local workspace.",
@@ -446,7 +473,7 @@ def responses_tool_definitions() -> list[dict]:
             "name": name,
             "description": descriptions[name],
             "parameters": TOOL_SCHEMAS[name],
-            "strict": True,
+            "strict": False,
         }
         for name in TOOL_NAMES
     ]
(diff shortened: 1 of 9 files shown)


hyeonsangjeon:
Blast radius of the v1 defect, checked rather than left implied: no experiment YAML in `batch-runner/experiments/` selects `execution_mode: agentic_sandbox` (0 of them). The mode is live and reachable from `core/executor.py`, so the overclaim is real, but nothing committed dispatches it today — it would first bite whoever writes a new experiment with that mode. That is the reason it can wait for a change that re-cuts `V1_TOOLS_SHA256` deliberately, rather than needing to ride along here.

Independent check of the fix, on the real path rather than the fixtures: the payload the paid probe will actually send was captured offline by driving `run_stage_a_probe` with a client that records and stops. Every key (`input`, `instructions`, `max_output_tokens`, `model`, `parallel_tool_calls`, `timeout`, `tools`) is one `Responses.create` accepts; both offered tools now carry `strict: false`; the tool definitions are the flat Responses shape with no keys outside `FunctionToolParam`; the input items carry no keys outside `EasyInputMessageParam`.

That rules out the structural causes of a `400` that can be ruled out offline. It does not prove the next dispatch succeeds — a service can refuse for reasons only it knows. What changed is that if it does, the record will name the status and the provider's code instead of just `BadRequestError`, and a refusal still costs nothing.

PR #490, opened 2026-09-10
feat(agentic-v2): stage B — a measured machine that could host the containment

hyeonsangjeon:
## What this is

Stage B of the Agentic Sandbox V2 activation asked exactly one question: **does a machine exist that could host the small isolated virtual machine the substrate manifest requires?** The answer is **yes**, measured on a real host rather than argued from documentation.

The plan for this stage was written and merged **before** the host was deployed (#489), so what follows can be checked against a prediction that was made in advance.

## The measurement

Deployed from `infra/dev-host` with no change to the template: `Standard_D8as_v5`, `koreacentral`, Ubuntu 24.04, `TrustedLaunch`, **no public IP and no inbound allow rule** — only the deny-all NSG rule, so the machine is reachable solely through the Azure control plane.

`scripts/check_agentic_containment.py --json` was run on it against a clone of this repo at `86152b7`. The report is sha256 `ee2222285af39a4674ed524865b08397413ab7d758543db20a8caf2477eda070`, verified against the VM's own computation of that hash after transfer.

| reading | result |
|---|---|
| processor virtualisation flag | `svm` — met |
| `/dev/kvm` | present, usable by an ordinary user — met |
| kernel release | `6.17.0-1022-azure`, above the 5.10 floor — met |
| `firecracker` / `jailer` | v1.13.1, both run — met |

## The plan's prediction was wrong

The merged plan expected the AMD default might not expose `/dev/kvm` and pre-authorised the Intel `Standard_D8s_v5`. The device is there. No size change was made and the pre-authorisation went unused. Nested virtualisation survived `TrustedLaunch` on this size, which does not hold everywhere, so it is recorded as a reading rather than carried forward as an assumption.

## What this deliberately does not change

`required_containment_available` is still `false`. The nine policy rules still read `cannot be established here`, because no module turns `REQUIRED_MICROVM_POLICY` into arguments for starting a virtual machine — that is stage C.

So exactly one aggregate moves:

- `could_be_hosted_on_any_machine_in_play` → `true`
- `available_on_any_machine_in_play` → stays `false`
- `refuse_command_execution` → still refuses, now naming the second ground rather than the first

No guard removed, no activation flag moved, `exec_run` still shut, no fallback to the V1 Docker runner.

## The two tests that pinned the old answer

Both were written to change on the day the answer changed, and both are updated rather than deleted:

- `test_no_recorded_finding_claims_the_containment_is_available` → `test_the_one_machine_that_could_host_it_does_not_claim_to_have_it`. A `True` in `could_host_the_containment` is the most misreadable value in the file, so the new test requires the finding to disclaim availability **in its own text**.
- The date assertion is loosened from one hardcoded day to an ISO date, which is what the test's own name promised.
- `test_the_whole_answer_covers_this_machine_and_the_recorded_ones` now asserts the aggregate is `true` while this machine cannot host it — which is what recording a finding is *for* — and that the second flag stayed false.

## Honest caveat

`firecracker` and `jailer` are not on the image; they were installed during the measurement. A freshly deployed host answers *no* on that fourth reading until bootstrapped. The three no install can supply — processor, device, kernel — are what make this a property of the machine.

## Cost

Allocated 08:40:50Z, deallocated by 08:54Z — about 13 minutes, confirmed `PowerState/deallocated`. Disks are kept and still charge. The per-hour rate under this subscription is not established here, so the entry is `partial`, **not zero**.

## Overlapping-file note

Touches `core/agentic_v2_containment_readiness.py`, its test, and the V2 task document — all B-owned V2 files. No shared executor, config, ledger, CI or CHANGELOG file is modified.

## Verification

- `pytest tests/test_agentic_v2_containment_readiness.py` — 64 passed
- All `tests/test_agentic*.py` — 1191 passed, 1 skipped, 8 deselected
- `mypy core/agentic_v2_containment_readiness.py` — clean; the 2 reported errors are in `core/agentic_v2_microvm.py`, untouched by this PR and pre-existing on `main`

### DIFF
--- a/batch-runner/core/agentic_v2_containment_readiness.py
+++ b/batch-runner/core/agentic_v2_containment_readiness.py
@@ -28,10 +28,16 @@ that report takes neither of them:
 It also reports only ``ready_for_boot_test`` or ``not_run``, which tells a reader
 that something is missing without telling them what or whether it is fixable.
 
-**Three machines are in play, and only one can be read from here.** The other
-two are recorded in :data:`RECORDED_FINDINGS` with the source and the date they
-were established, kept deliberately separate from anything probed, so a reader
-is never left guessing which kind of evidence they are looking at.
+**Four machines are in play, and only one can be read from here.** The other
+three are recorded in :data:`RECORDED_FINDINGS` with the source and the date
+they were established, kept deliberately separate from anything probed, so a
+reader is never left guessing which kind of evidence they are looking at.
+
+One of those three, the Azure dev host, is the first to answer yes to the
+hosting question. That moved one of the two questions this module asks and left
+the other exactly where it was: no code applies these rules, so the containment
+is still in place nowhere, and the refusal below still stands on that second
+ground.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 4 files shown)


hyeonsangjeon:
Pushed a third commit after re-reading the branch against the repo rather than against my own summary.

`tests/test_agentic_v2_containment_rules.py` opened by giving **two** reasons the boot-and-exceed test cannot be written yet: no machine in play can host the containment, and no code turns the rules into arguments for starting one. The finding added by this branch makes the first one false — so without the fix, this PR would have shipped a file asserting something the same PR disproved.

Now one reason, named as one. The file's own guard (`assert "it cannot be written yet" in __doc__`) did not catch it because it pins the conclusion, not the reasoning underneath — worth knowing for when C3 finally deletes that sentence.

Also checked: this was the only stale claim. The other two matches in the tree are the conditional branch text in `refuse_command_execution` (still correct, just no longer the branch taken) and the test that asserts the refusal *doesn't* say it, which this change makes more relevant rather than less.

PR #491, opened 2026-09-10
docs(agentic-v2): stage C plan, written before any of it is built

hyeonsangjeon:
Stage B ended with a machine that *could* host the containment and no code that
applies it. This records what stage C will do — its conditions and its failure
responses — before the first line of it is written. That ordering is what made
stage B's wrong prediction visible rather than quietly amended, and it is the
only reason this PR exists separately from the code it plans.

**Documentation only.** One file, no code, no test, no flag. `exec_run` stays
shut, `foundation_only` and `production_activation: disabled` are untouched, and
no guard is removed.

### Four changes, not three — because writing the plan found a gap

The seventh attack in the existing stage C spec is that a command must not be
able to read a token, key or environment secret the orchestrator holds. Checking
the sources rather than assuming: **the words *credential*, *secret* and *token*
do not appear** in `core/agentic_v2_substrate.py`, in
`security/agentic-v2-supply-chain-policy.json`, or in
`sandbox/agentic_v2_capabilities.json`. The containment answers six questions —
where a command may write, network, memory, duration, who it runs as, what
happens on breach — and the credential boundary is not among them.

So that attack as specified would have tested a boundary no rule states. That is
the same pretend-enforcement stage C exists to end, only pointed the other way:
instead of a rule nothing applies, a test with nothing behind it. **C0** writes
the rule down first; C3's attack then aims at something real.

Scope of C0 measured rather than guessed: `_SUPPLY_CHAIN_RULE_NAMES` translates
only four of the ten containment rules into the signed policy's naming, and the
other six — every limit with a number — live on the containment side alone. A
seventh joins them there, deliberately **not** added to the signed policy, since
that file carries a signature and extending it is a separate question.

### What else is read off primary sources

- **The eleven policy keys**, each mapped to where it actually lands in
  Firecracker's configuration. `network: none` is asserted as *no interface
  configured*, because Firecracker's own fixtures write it `"network-interfaces":
  []` and a test that only knows the absent-key form can be walked around.
- **Four devices outside the policy that each defeat a rule inside it** —
  `mmds-config`, `vsock`, `memory-hotplug`, `balloon` — all required `null`. They
  are not added to `REQUIRED_MICROVM_POLICY`: they are not rules about what the
  containment allows, they are devices that must be absent for its existing rules
  to mean what they say.
- **The jailer runs as root by its own documented design**, because building the
  jail needs `mknod`, `pivot_root` and cgroup writes, and drops to `--uid`/`--gid`
  immediately before exec. Written down so nobody later reads that root in `ps`
  as a rule quietly dropped. No root for the model's command, no `--privileged`,
  no capability added to make a test pass.
- **The supply-chain policy governs OCI images, not VM images.** Its
  `required_evidence` names `oci_layout` and its provenance subjects include
  `dockerfile`. A `vmlinux` has no parent in that chain, so C2 records it as
  provenance **weaker than the OCI chain's**, in those words, rather than listing
  a hash and letting the reader assume it was signed. The rootfs keeps the chain
  by being built from the image the policy already governs —
  `core/agentic_v2_oci.py` verifies that layout blob by blob and its layers are
  `application/vnd.oci.image.layer.v1.tar`.
- **The kernel/rootfs seam already exists.** `inspect_microvm_readiness` takes
  `asset_paths`, hashes both, and gates `ready_for_boot_test` on them. The
  launcher reuses those hashes rather than computing its own, so there is one
  answer to "which kernel booted".

### Why C2 is separated from C1

This box runs kernel 3.10 and cannot boot a Firecracker machine at all. C0 and
C1 are checkable here and cost nothing; C2's boot is testable only on the Azure
dev host stage B measured. Keeping them apart is what stops the second being
claimed on the strength of the first.

### Verification

Full backend suite on the parent commit: **9531 passed, 12 skipped, 45
deselected** (21m24s, exit 0). This PR changes no code, so that stands. Nothing
in `tests/` or `core/` reads this document.

The dev host was re-checked while writing this and is still
`PowerState/deallocated` — no silent billing since stage B.

### DIFF
--- a/tasks/0822_saturday/TASK_AGENTIC_SANDBOX_V2_ACTIVATION.md
+++ b/tasks/0822_saturday/TASK_AGENTIC_SANDBOX_V2_ACTIVATION.md
@@ -827,9 +827,31 @@ policy holds. The weaken test does not — it is parametrised by hand, so C0 add
 its entry there too, a coverage gap rather than a failure and therefore the kind
 that stays open unless it is written down.
 
+**Six becomes seven, and prose does not follow a dict on its own.** The last
+change of this shape made a docstring false — PR #490 added a finding and left a
+test file still asserting the opposite, caught only by grepping the tree
+afterwards. The same failure is available here, so the places that say *six* are
+listed before the change rather than hunted after it:
+
+| says six | change it |
+|---|---|
+| `core/agentic_v2_substrate.py:143` — "Six things have to be stated" | yes |
+| `core/agentic_v2_substrate.py:222` — "one of the six questions" | yes |
+| `tests/test_agentic_v2_containment_rules.py:3` — "Six questions have to be answered" | yes |
+| `tests/test_agentic_v2_containment_rules.py:71` — section comment | yes |
+| the test named `test_every_one_of_the_six_questions_has_an_answer` | yes, renamed |
+| this document, the inherited-state table and stage C's exit condition | yes |
+| `CHANGELOG.md` and `TASK_AGENTIC_SANDBOX_V2_FOUNDATION.md` | **no** |
+
+The last row is the one worth stating. Those record what was true on
+2026-08-26, when six was the answer. Rewriting them would not fix a stale
+sentence, it would falsify a record of when the rule set changed — and the
+count moving is exactly the thing a history is for.
+
 **Exit condition.** The new rule is stated, the manifest equals the policy again,
-weakening it is refused like the other ten, and `containment_rules_that_disagree`
-still returns nothing.
+weakening it is refused like the other ten, `containment_rules_that_disagree`
+still returns nothing, and nothing in `core/` or `tests/` still says the
+containment answers six questions.
 
 **Failure response.** If the boundary cannot be stated as a rule the launcher
 could apply, C3's attack 7 is recorded as untestable with the reason, rather than
--- a/tasks/0822_saturday/TASK_AGENTIC_SANDBOX_V2_ACTIVATION.md
+++ b/tasks/0822_saturday/TASK_AGENTIC_SANDBOX_V2_ACTIVATION.md
@@ -884,6 +884,23 @@ the policy describes even though no key names them: `--new-pid-ns`, so the guest
 process is not in the host's PID namespace, and `--cgroup` for the host-side
 memory bound that sits underneath the guest-visible one.
 
+**And a third, which is a default that would quietly be wrong here.**
+`--cgroup-version` defaults to `1` in the jailer's own documentation, while
+Ubuntu has defaulted to the cgroup **v2** unified hierarchy since 21.10 and the
+host stage B measured runs 24.04. A launcher that passes `--cgroup` and leaves
+the version at its default would be writing to a hierarchy that is not the one
+in use — a memory bound that does not apply, on a host where nothing would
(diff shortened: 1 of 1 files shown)


hyeonsangjeon:
Second commit corrects the PR body's C0 scope: **three places, not two.**

The plan said the credential rule lands in `REQUIRED_MICROVM_POLICY` and
`sandbox/agentic_v2_capabilities.json`. Rather than leave that as a reading, I
added `credentials: "none-inherited"` to both on a scratch copy and ran every
agentic test. One failure:
`test_a_machine_with_everything_still_does_not_have_the_containment`
(1 failed, 1262 passed, 3 skipped).

The readiness report builds one sentence per rule from
`_POLICY_SETTING_AS_A_CLAIM`. A key with no wording there does **not** pass
silently — it emits a line saying the manifest "has gained a containment setting
this report does not know how to describe or check", which is not the
"unenforced rather than met" the test requires. So the third place is
`core/agentic_v2_containment_readiness.py`, and the codebase names it itself.

What did *not* break is the more useful half. `containment_rules_that_disagree`
and the signed-policy validation both passed untouched — confirming the plan's
claim that the rule does not need to reach the signed file. The drop test and
the manifest-equality test picked the new key up generically, as predicted. The
weaken test did not, because it is parametrised by hand: a coverage gap rather
than a failure, and therefore exactly the kind that stays open unless written
down, so C0 adds that entry too.

Scratch edits reverted; `git status` clean. This commit changes documentation
only.

hyeonsangjeon:
Three more corrections, all from checking the plan's claims against primary
sources rather than shipping them as written.

**`--cgroup-version` defaults to `1`, and this host is cgroup v2.** Checked every
flag the C1 mapping passes against the jailer's own documentation. The spellings
all hold — `--exec-file`, `--uid`, `--gid`, `--netns`, `--new-pid-ns`, `--cgroup
<file>=<value>`, `--resource-limit fsize=<bytes>`. What the check turned up is a
default the plan had not accounted for: a launcher that passes `--cgroup` and
inherits `--cgroup-version=1` writes to a hierarchy Ubuntu has not used since
21.10. **A memory bound that silently does not apply** — this stage's own failure
mode arriving through a default instead of a deletion. Passed explicitly now.

The v2 half is a fact about the distribution and *not* a reading from that
machine; stage B measured the processor, `/dev/kvm`, the kernel and the two
programs, and not this. C2 reads it off the host first, and the plan says so
rather than letting an inference be mistaken later for a measurement.

**Checked against v1.13.1, not `main`.** The first check hit the development
branch, which can document a flag the deployed binary does not have. Repeated
against the v1.13.1 docs — the version stage B found installed. `--cgroup-version`
is present there and defaults to `1`, so the point above holds for the binary
that will actually run.

**Seven places say "six questions"; C0 makes it seven.** This is the shape of
change that went wrong in #490 — a finding added, a test file left asserting the
opposite, caught only by grepping the tree afterwards. So the places are listed
before the change rather than hunted after it.

The row worth reading is the one marked **no**: `CHANGELOG.md` and
`TASK_AGENTIC_SANDBOX_V2_FOUNDATION.md` record what was true on 2026-08-26, when
six was the answer. Rewriting those would not fix a stale sentence — it would
falsify the record of when the rule set changed, which is what a history is for.

Still documentation only. No code, no test, no flag; `exec_run` stays shut.

PR #494, opened 2026-09-10
feat(agentic-v2): stage C0 — the containment's seventh question, found by planning its test

hyeonsangjeon:
## What this is

Stage C0 of the Agentic Sandbox V2 activation plan merged in #491. It is the smallest change in stage C and the one that had to come first, because C3's seventh attack was aimed at a boundary no rule stated.

## Why

Stage C's attack list ends with one unlike the others: *the command must not be able to read a token, key or environment secret the orchestrator holds.* Writing it out properly meant asking which rule it enforces — and the answer was none. At `0076295`, the commit this was written against, the words `credential`, `secret` and `token` appeared **nowhere** in `core/agentic_v2_substrate.py`, `security/agentic-v2-supply-chain-policy.json` or `sandbox/agentic_v2_capabilities.json`: zero matches in all three.

That test would have passed against nothing at all. **A test with no rule behind it is the same fault as a rule with no code behind it, seen from the other side** — and this repository has spent a lot of effort on the first kind.

## What changed

`credentials: "none-inherited"` is now the eleventh containment rule, in `REQUIRED_MICROVM_POLICY` and in the manifest that travels with a built image, which the tests require to be **equal**, not merely compatible.

- weakened to `inherit-environment` → refused
- softened to a bare `"none"` → refused
- deleted → refused
- readiness report carries it beside the other ten as *"cannot be established here"* — the same verdict, for the same reason: nothing turns any of these rules into arguments for starting a machine yet

Every other rule here arrived because something was written down and nothing applied it. This one arrived the opposite way.

## What checking it against the jailer changed

The plan said the launcher could lean on the jailer, which "already clears inherited environment variables and file descriptors before exec". Firecracker's jailer documentation for **v1.13.1** — the version stage B measured on the Azure host, not the development branch — says something narrower:

> "cleanup all environment variables received from the parent process"

> "close all open file descriptors based on `/proc/<jailer-pid>/fd` **except input, output and error**"

Those three survive, and are pointed at `/dev/null` only by the separate step that runs when the launcher asks to daemonize. So the rule holds **by default** for the environment and only **conditionally** for the descriptors.

Consequences, written into the plan rather than discovered later:

- `--daemonize` moves from something C1's launcher may pass to something it **has to**
- C3's attack 7 inspects the descriptors as well as the environment, because an attack checking the environment alone would pass on a launcher that had dropped the flag
- `--daemonize` sends Firecracker's stdout to `/dev/null`, which is acceptable only because results come off the ephemeral work disk, not the console — stated in C1 so it is not met as a bug

A rule written down on the strength of a mechanism that turns out to work differently is the fault this stage exists to fix, one layer further down.

## One thing the check turned up that C1 and C3 both need

The signed supply-chain policy is **not** a second copy of the containment. It mirrors four rules — `runtime`, `network`, `read_only_rootfs`, `ephemeral_work_disk` — and always has. The other seven, including all three numeric limits, `on_breach`, `user` and now `credentials`, live in `REQUIRED_MICROVM_POLICY` and the manifest only.

Nothing is wrong with that and it is not changed here. It is recorded because the drift check between the two files can only ever guard the four, so *"the signed policy still validates"* is not the same statement as *"the containment is intact"*, and a launcher built by reading the signed policy would enforce well under half of it.

## What was deliberately not renumbered

Three records still say six or nine:

| record | changed |
|---|---|
| `CHANGELOG.md`'s 2026-08-26 entry | no |
| `TASK_AGENTIC_SANDBOX_V2_FOUNDATION.md` (five places, all dated) | no |
| the recorded Azure finding in `core/` | clause added, number kept |

Only the last needed one, because it sits beside a report a reader can run today and get ten from. Rewriting the two dated documents would not fix a stale sentence; it would falsify a record of when the rule set changed, which is what those records are for.

The two rows of the activation plan's section 1 that stage B and C0 overtook **on the same day** say so in the row rather than being rewritten — a table headed *"read from the code on 2026-09-10"* cannot silently mean two different states of that day.

## Shared-file note

`CHANGELOG.md` is shared with the Codex work. This adds three bullets to `Unreleased / Added` **below** the existing Codex entries and modifies none of them. If the other side merges first, this rebases onto it cleanly.

## Verification

- `tests/test_agentic_v2_containment_rules.py` — 35 passed
- agentic / containment / substrate / supply-chain / microvm selection — **1,280 passed, 3 skipped**
- full backend suite — **9,559 passed, 12 skipped, 46 deselected, exit 0** (23m46s)
- readiness report on this box: 14 requirements → 15, 12 missing → 13, the new one reading *"the command's access to the tokens, keys and environment the orchestrator holds is none-inherited — cannot be established here"*

## What this is not

No guard removed, no policy weakened, no fallback to V1's Docker backend, no activation flag flipped, no `exec_run` opened, nothing run as root, nothing spent. `foundation_only` stays `true` and `production_activation` stays `disabled`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -90,6 +90,65 @@ entries land under a fresh dated heading the day they merge to `main`.
   runner. The gate stays closed; only the reason is now true.
 
 ### Added
+- **The containment answers a seventh question, and it was found by planning
+  the test rather than by reading the rules.** Stage C's attack list has always
+  ended with one that is not like the others: the command must not be able to
+  read a token, key or environment secret the orchestrator holds. Writing it out
+  properly meant asking which rule it enforces — and the answer was none. At
+  `0076295`, the commit this was written against, `credential`, `secret` and
+  `token` appeared nowhere in `core/agentic_v2_substrate.py`,
+  `security/agentic-v2-supply-chain-policy.json` or
+  `sandbox/agentic_v2_capabilities.json`: zero matches in all three. The test
+  would have passed against nothing at all.
+
+  **A test with no rule behind it is the same fault as a rule with no code
+  behind it, seen from the other side**, and this repository has spent a lot of
+  effort on the first kind. So `credentials: "none-inherited"` is now the
+  eleventh containment rule, in `REQUIRED_MICROVM_POLICY` and in the manifest
+  that would travel with a built image, which the tests require to be equal
+  rather than merely compatible. It is
+  refused if it is weakened to `inherit-environment`, refused if it is softened
+  to a bare `"none"`, refused if it is deleted, and the readiness report now
+  carries it beside the other ten as *"cannot be established here"* — the same
+  verdict, for the same reason, because nothing turns any of these rules into
+  arguments for starting a machine yet.
+
+  It does **not** appear in the signed supply-chain policy, and that is not an
+  omission: that file mirrors four of the eleven rules — `runtime`, `network`,
+  `read_only_rootfs`, `ephemeral_work_disk` — and always has. Seven rules,
+  including the three numeric limits and `on_breach`, live in the policy and the
+  manifest only. Worth knowing before someone reads the signed policy as the
+  full statement of the containment; it is a signed subset of it.
+
+  Every other rule here arrived because something was written down and nothing
+  applied it. This one arrived the opposite way: nothing was written down, and a
+  test was about to be aimed at it anyway.
+
+- **Checking that rule against the jailer's own documentation moved a flag from
+  optional to required.** The plan said the launcher could lean on the jailer,
+  which "already clears inherited environment variables and file descriptors
+  before exec". Firecracker's jailer documentation for **v1.13.1** — the version
+  measured on the Azure host, not the development branch — says something
+  narrower: it will *"cleanup all environment variables received from the parent
+  process"*, but *"close all open file descriptors … **except input, output and
+  error**"*. Those three survive, and get pointed at `/dev/null` only by the
+  separate step that runs when the launcher asks to daemonize.
+
+  So the new rule holds by default for the environment and only conditionally
+  for the descriptors. `--daemonize` is now something stage C1's launcher has to
+  pass rather than something it may, and C3's attack on this rule inspects the
+  descriptors instead of trusting the sentence. A rule written down on the
+  strength of a mechanism that turns out to work differently is the fault this
+  stage exists to fix, one layer further down.
+
+- **Three records still say six, and are left saying it.** `CHANGELOG.md`,
+  `TASK_AGENTIC_SANDBOX_V2_FOUNDATION.md` and the recorded Azure finding all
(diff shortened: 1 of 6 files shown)


hyeonsangjeon:
## Second commit: two things C1 would have built on, corrected

Found while this PR's CI ran, by reading the sources the plan cites rather than the plan. Docs only — the code is unchanged and its full run was green (`pytest` 23m11s, `advance-check` 2m2s, `freeze-check` 18s).

### 1. Firecracker has no memory hotplug

The plan's device table said **four** devices a launcher must pin at `null` or lose a rule it appears to enforce. The fourth was `memory-hotplug`, on the reasoning that memory added after boot is memory the bound never saw. The reasoning is sound; the device is not.

- v1.13.1's own configuration fixture has no such key
- its API specification — 43 KB, every route — contains **no occurrence of `hotplug`**
- the routes are `/actions`, `/balloon`, `/balloon/statistics`, `/boot-source`, `/cpu-config`, `/drives`, `/entropy`, `/logger`, `/machine-config`, `/metrics`, `/mmds`, `/mmds/config`, `/network-interfaces`, `/snapshot/create`, `/snapshot/load`, `/version`, `/vm`, `/vm/config`, `/vsock`

Balloon **is** the runtime-memory mechanism in this Firecracker. The table had it twice under two names, and a test pinning `memory-hotplug: null` would have asserted something about a key Firecracker never emits — and passed for that reason.

**That is the third check in this stage that would have held against nothing**: attack 7 against a credential rule that did not exist, the jailer's three surviving descriptors, and this one.

Added in its place, as a route rather than a device: **`/snapshot/load`** restores a machine whose configuration was decided elsewhere, so it can carry a NIC, a vsock or a different memory size that the launcher's own arguments would never show. It is not a setting to pin at `null`; it is a way in that bypasses the table entirely, so C3 treats loading one as an escape rather than a configuration.

The keys the plan *did* get right were re-verified against the same fixture and stand: `is_read_only`, `is_root_device`, `machine-config.mem_size_mib`, `network-interfaces: []`, `vsock: null`, `balloon: null`, `mmds-config: null`.

### 2. A fourth partial copy of the containment

`core/agentic_v2_microvm.py` reports `network`, `rootfs_mode` and `workdir` at the top level of its readiness report as the string literals `"none"`, `"read-only"` and `"ephemeral-quota"`. It imports `canonical_sha256` from the substrate and nothing else — **not** `REQUIRED_MICROVM_POLICY` — and `validate_microvm_readiness_report` checks the report against the same literals rather than against the rules.

Not a hole that hides anything today: all three are among the four the signed policy mirrors, so weakening one still fails `containment_rules_that_disagree` — the stale report would appear *beside* a failure rather than instead of one.

Left alone here because C0's scope is a missing rule and not a drift surface, and because the fix belongs to the module that will read the policy for real. **C1 imports `REQUIRED_MICROVM_POLICY`, restates no value of it, and takes this one with it rather than adding a fifth.**

PR #497, opened 2026-09-10
feat(agentic-v2): build the launch arguments the containment rules ask for

hyeonsangjeon:
## What this is

Stage C1 of the Agentic Sandbox V2 activation plan. Stage C0 left eleven containment rules written down in `REQUIRED_MICROVM_POLICY` and **nothing anywhere turning any of them into an argument**. This adds `core/agentic_v2_microvm_launch.py`, which reads the policy and returns the jailer and Firecracker arguments that would apply each rule.

It **starts nothing** — no process, no machine, no `exec_run`. That is what makes it checkable on a box whose 3.10 kernel could not boot a microVM, and it is why the report below still says the containment is not in place.

## The field that was nearly wrong

The readiness report gains a fourth field, `every_rule_has_an_argument_that_would_apply_it`, derived from the builder against the policy rather than declared, so a rule added to one and not the other turns it `False` on its own.

`anything_applies_the_containment_rules` **stays `False`**, and the new field is deliberately **not** an input to `available_on_any_machine_in_play`. The obvious wiring would have been wrong: a recorded finding already carries `could_be_hosted_anywhere`, so a `True` here would have flipped the aggregate and made the report announce the containment as **in place** on the strength of code that has never booted anything. That is the exact failure the readiness module exists to prevent, arriving through good news instead of through a deletion.

## Two corrections to the merged plan

Every flag was checked against **v1.13.1** — the version stage B found installed — not the development branch. Two readings contradicted the plan text, and both are corrected in this PR:

- **`--no-api` is required, and "emits no snapshot route" was too weak.** Pinning `balloon`, `vsock` and `mmds-config` out of a config file is a statement about *start-up*. With the API socket live it says nothing about the run: all three can be added back after boot, and `/snapshot/load` can bring in a machine configured somewhere else entirely. `--no-api` (`src/firecracker/src/main.rs:198-205`, takes no value, requires `--config-file`) is what makes the configuration file the whole of what the machine will ever be.
- **`--new-pid-ns` does not write the PID file.** The plan said it did. `src/jailer/src/env.rs:735-740` writes it in **both** branches, after `chroot()`, at the in-jail path `/firecracker.pid`. Both flags are still passed for their own reasons; only the explanation was wrong.

## Two gaps named rather than papered over

- **No rule about processor share.** Firecracker requires `vcpu_count` and the policy has none, so it comes from the caller and the gap is written down. Closing it means amending `REQUIRED_MICROVM_POLICY` in a change of its own, the way the credential rule was added — not defaulting a number in a launcher.
- **`/dev/net/tun` is in the jail regardless.** The jailer `mknod`s it unconditionally, so `network: none` rests on no interface being configured and `--netns` being absent, not on the device being missing. C3 should expect to find the node there.

## How the tests were checked

81 tests, and they were checked by **breaking the builder**, not by passing. Fourteen deliberate mutations — the flag deleted, the host memory bound dropped to the guest's, the v1 cgroup file name used under v2, the root drive made writable, the in-jail config path turned into a host path — each had to fail a test.

**Two initially did not, and both were holes in the tests rather than the builder:**
- `fsize=` was asserted as a string anywhere in the argument list, so deleting the `--resource-limit` that carries it changed nothing.
- The in-jail paths were asserted against their own constants, so moving a constant to a host path took the test with it.

Both now assert the requirement instead of the spelling. The last survivor was the builder's own backstop — the check that every accepted rule reached `rules_applied`, which no healthy build exercises — and it now has a test that stages the drop it exists for.

## Verification

- Full backend suite: **9654 passed, 12 skipped**, exit 0 (21m09s).
- `mypy` clean on the new module. The two pre-existing errors in `agentic_v2_microvm.py` are unchanged from `HEAD` and are in the backlog `backend-tests.yml` documents itself as not gating on.

## Shared-file note (per the A/B ownership split)

- `CHANGELOG.md` — new entry added at the **top of `### Added`**; none of A's Codex entries under `### Changed` / `### Fixed` are touched.
- No change to `executor.py`, shared config, the ledger, or shared CI.

## What this is not

Not an activation. `foundation_only` and `production_activation` are untouched, no guard was removed, no policy weakened, no fallback to V1's Docker runner, and `exec_run` stays closed. **Environment preparation, not 220-task execution.** C2 (first boot on the Azure host, reading the cgroup version off it as its first action) and C3 (the seven escape attempts) are still ahead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -228,9 +228,22 @@ entries land under a fresh dated heading the day they merge to `main`.
     makes the configuration file the whole of what the machine will ever be.
   - **`--new-pid-ns` does not write the PID file.** The plan said it did.
     `src/jailer/src/env.rs:735-740` writes it in both branches, after `chroot()`,
-    at the in-jail path `/firecracker.pid`. Both flags are still passed; only the
+    so the file lands inside the jail. Both flags are still passed; only the
     explanation was wrong, and it is the kind of wrong that surfaces as
     confusion at the moment somebody omits the flag and finds the file anyway.
+
+    Checking that correction turned up a **bug in the new module**, which is the
+    part worth recording. The PID file is named after the **exec file**, not
+    after the word: `save_exec_file_pid` appends `.pid` to the binary's own
+    name, and the jailer requires that name to *contain* `firecracker` rather
+    than to be it. So `/opt/firecracker-v1.13.1` writes
+    `/firecracker-v1.13.1.pid`, and the hard-coded `/firecracker.pid` would have
+    named a file that never appears on any host whose binary carries a version
+    suffix. Nothing would have failed — the deadline would have been recorded,
+    the plan would have hashed, and `wall_clock_seconds` would have had nothing
+    to act on at the moment it was needed. Derived from the binary now, with a
+    test that asserts it against a versioned name rather than against the
+    constant it used to be pinned to.
 
   Two gaps are named rather than papered over. The policy has no rule about
   processor share, so `vcpu_count` comes from the caller and the missing rule is
@@ -240,17 +253,19 @@ entries land under a fresh dated heading the day they merge to `main`.
   on `--netns` being absent, not on the device being missing; stage C3 should
   expect to find the node there.
 
-  81 tests, and they were checked by breaking the builder rather than by
-  passing. Fourteen deliberate mutations — the flag deleted, the host memory
+  83 tests, and they were checked by breaking the builder rather than by
+  passing. Fifteen deliberate mutations — the flag deleted, the host memory
   bound dropped to the guest's, the v1 cgroup file name used under v2, the root
-  drive made writable — each had to fail a test. Two initially did not, and both
-  were holes in the tests: `fsize=` was matched as a string anywhere in the
-  argument list, so removing the `--resource-limit` that carries it changed
-  nothing, and the in-jail paths were asserted against their own constants, so
-  moving one to a host path took the test along with it. Both now assert the
-  requirement rather than the spelling. The last survivor was the builder's own
-  backstop, which no healthy build exercises, and it now has a test that stages
-  the drop it exists for.
+  drive made writable — each had to fail a test. Three initially did not, and
+  all three were holes in the tests rather than in the builder: `fsize=` was
+  matched as a string anywhere in the argument list, so removing the
+  `--resource-limit` that carries it changed nothing; the in-jail paths were
+  asserted against their own constants, so moving one to a host path took the
+  test along with it; and the PID file was asserted against the constant that
+  turned out to be wrong, so it agreed with the bug rather than catching it.
+  All three now assert the requirement rather than the spelling. The last
+  survivor was the builder's own backstop, which no healthy build exercises, and
+  it now has a test that stages the drop it exists for.
 
 - **The containment answers a seventh question, and it was found by planning
   the test rather than by reading the rules.** Stage C's attack list has always
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -267,6 +267,21 @@ entries land under a fresh dated heading the day they merge to `main`.
   survivor was the builder's own backstop, which no healthy build exercises, and
   it now has a test that stages the drop it exists for.
 
+  One more thing broke on the way out, and it was prose rather than code.
+  Reading the generated report afterwards showed it asserting both halves of a
+  contradiction three lines apart: the new section said every rule now has an
+  argument, while the recorded azure finding above it still ended by saying no
+  code turns the policy into launch arguments. That sentence was true the day it
+  was measured and false the moment this module merged, and **nothing failed**,
+  because a recorded finding is a frozen string and no test compared it against
+  the live answer. The fix is structural rather than editorial: whether anything
+  applies the rules is a property of this repository, it changes without any
+  machine changing, and `describe_containment` already answers it live in its
+  own section — so a finding may no longer name `REQUIRED_MICROVM_POLICY` or the
+  launch module at all, and a test enforces that for every finding rather than
+  for the one that went stale. The same reach appeared in a test module's
+  docstring and is corrected there too.
+
 - **The containment answers a seventh question, and it was found by planning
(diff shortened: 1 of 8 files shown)


hyeonsangjeon:
## Update: independent re-verification found a bug in this PR's own code

The two v1.13.1 corrections above are load-bearing — they change a merged plan document — so I re-checked both against the upstream source rather than against the reading that produced them.

**`--no-api` confirmed** as described: `Argument::new("no-api").takes_value(false).requires("config-file")`, and `let api_enabled = !arguments.flag_present("no-api")`.

**`--new-pid-ns` confirmed** as described — `save_exec_file_pid` runs in both branches, after `chroot()`. But re-reading it surfaced something the first pass had missed, and it was a bug in this PR:

> The PID file is named after the **exec file**, not after the word. `save_exec_file_pid` appends `.pid` to `chroot_exec_file`, which is `/` joined to the binary's own name — and the jailer requires that name to *contain* `firecracker`, not to be it.

So `/opt/firecracker-v1.13.1` writes `/firecracker-v1.13.1.pid`. This PR had it hard-coded as `/firecracker.pid`. On any host whose binary carries a version suffix, the plan would have named a file that never appears.

**Nothing would have failed.** The deadline would have been recorded, the plan would have hashed cleanly, and `wall_clock_seconds` would have had nothing to act on at the moment it was needed — a limit that reads as enforced and is not, in the one field that decides whether the clock rule can be enforced at all. That is precisely the failure mode stage C exists to end, and it was sitting in stage C's own code.

**The test agreed with the bug**, because it asserted the path against the same constant — the third instance in this PR of a test pinned to a spelling rather than to a requirement. It now asserts against a versioned binary name and fails if the constant comes back; I verified that by reintroducing the bug and watching both parametrised cases fail.

Fixed in `882b436`. 83 tests now; agentic suites 1003 passed, 2 skipped. The plan document and CHANGELOG both record the finding, since the value of it is the pattern rather than the two-line fix.

hyeonsangjeon:
### Third commit: `dda2b5d` — a stale sentence the second commit's own good news made false

Reading `describe_containment`'s output after the fact, rather than reading the tests, turned up the report asserting both halves of a contradiction three lines apart:

> **azure dev host** … This says the containment could be hosted, and nothing more: **no code turns `REQUIRED_MICROVM_POLICY` into arguments for starting a virtual machine**, so on this machine too the rules are written down and unapplied

> **Whether anything applies the rules, on any machine**
> **Every rule now has an argument that would apply it:** `core.agentic_v2_microvm_launch` builds them.

The first was true on 2026-09-10 when the host was measured and false the moment this PR's first commit merged into the branch. **Nothing failed.** A recorded finding is a frozen string, and no test compares it against the live answer — the same shape as the PID-file bug in `882b436`, where a wrong constant and a test pinned to that constant agreed with each other.

**The fix is structural, not a reword.** Whether anything applies the rules is a property of *this repository*: it changes without any machine changing, and `describe_containment` already answers it live in its own section. So a finding may no longer name `REQUIRED_MICROVM_POLICY` or the launch module at all, enforced for every finding rather than patched on the one that went stale:

```python
def test_no_finding_answers_the_repository_wide_question_for_itself():
    answers_that_are_not_about_a_machine = (
        "REQUIRED_MICROVM_POLICY",
        "agentic_v2_microvm_launch",
    )
```

`tests/test_agentic_v2_containment_rules.py`'s module docstring had reached for the same answer ("what is still missing is the other half, that no code turns these rules into arguments") and is corrected: the missing half is now the launch, not the translation.

Two smaller things in the same pass, both about the section this PR added:

- The second line printed `NOTHING_APPLIES_THESE_RULES_YET` verbatim. That sentence is worded for one rule's verdict, so under a section heading it says "this rule" with no antecedent and restates the line above it. Replaced with the section's own words; the constant is still printed once per rule further down (12 occurrences).
- The test for the section matched `"Nothing does:"` against the **whole report**, which would have kept passing after the line was deleted, because the same sentence appears per rule. It now slices the section out by heading and asserts inside it. Verified by deleting the line: `assert "starts no process" in section` fails as it should.

**Verification.** 1366 passed / 3 skipped across `-k "agentic or containment or substrate"`; full backend suite running. CI on `882b436` was green on all three checks (`advance-check`, `freeze-check`, `pytest` 23m28s) before this push.

**Still `False`:** `anything_applies_the_containment_rules`. Nothing in this PR starts a process.

hyeonsangjeon:
### Fourth commit: `5bd31af` — the fix in `dda2b5d` was right about the printout and wrong about the JSON

`dda2b5d` had the azure finding hand the repository-wide question to *"the section below"*. That is true of `describe_containment`'s printout and false of the other way this report is read — `scripts/check_agentic_containment.py --json` emits the same string as a JSON value, where there is no section and no below. The finding now names `anything_applies_the_containment_rules`, which is correct in both.

**Naming a field trades one quiet failure for another, so it has its own guard.** A stale *claim* is at least visible to a reader who knows the answer changed. A stale *pointer* is visible to nobody: rename the key and the sentence still reads perfectly while referring to nothing. So anything in a finding shaped like a field name has to be one:

```python
looks_like_a_field = re.compile(r"[a-z][a-z0-9]*(?:_[a-z0-9]+)+")
...
assert field in report, (
    f"the {finding.machine} finding points at {field}, which is not a key "
    "of the report; a pointer that no longer resolves is the staleness "
    "this rule exists to stop"
)
```

Sixteen characters with an underscore. Across all three findings that threshold matches exactly one token today — the key I put there — and nothing in the surrounding prose.

**Both halves checked by breaking them, not by their passing:**

| mutation | result |
|---|---|
| put `REQUIRED_MICROVM_POLICY` back into the finding | `test_no_finding_answers_the_repository_wide_question_for_itself` fails |
| rename the report key out from under the finding | `test_a_finding_that_points_at_a_report_field_points_at_a_real_one` fails, naming the finding and the missing key |

**Verification.** 1367 passed / 3 skipped on `-k "agentic or containment or substrate"`; full backend suite restarted on the settled tree — the earlier run overlapped the mutation experiments, which rewrite a source file in place, so its result was not usable evidence and was discarded rather than reported.

`anything_applies_the_containment_rules` is still `False`. Nothing in this PR starts a process.

PR #510, opened 2026-09-10
feat(agentic-v2): D3b — sweep the declared manifest, and the runner that spends stage D

hyeonsangjeon:
Two things, both of which had to happen before stage D is allowed to cost anything.

## The overclaim, closed

`run_stage_d_probe`'s docstring and `ProbeCannotDescribeItself` both read as though naming the substrate manifest verified the guest image. It does not. The manifest is a **declaration** — twenty commands shaped `{"name": "Rscript", "probe": ["Rscript", "--version"]}`, no versions, no hashes — plus a `supply_chain` block naming *policy profiles* (`cosign-offline-v1`) rather than the result of applying them.

Worse, there is a gap underneath it that the repository was not saying out loud:

| | |
|---|---|
| the receipt that reads `verified` | measured on `sha256:e47537b8…`, a candidate **never pushed** and not present on any host here |
| the digest stage D is handed | `sha256:ee6ef798…` — which `parent.lock.json` pins, i.e. the **parent** the candidate was built *from* |
| signature / provenance / CVE | `not_run` for both, per `sandbox/v2/README.md` |

So the image stage D can fetch is not the image the capability evidence describes.

`capability_evidence_for()` reads that off the lock file rather than asserting it, in the vocabulary `core/agentic_v2_supply_chain.py:57` already uses — `verified` / `failed` / `not_run`, plus `unknown` for a digest nothing here holds a record of. The answer travels in the record as `image_evidence`, so a stage D result can never be read as a claim that the image it booted was verified. For the fetchable digest the answer is `not_run`, and that is the point of having the field.

A sha256 this repository computed over bytes it fetched is an identity for those bytes. It is not a supplier's assertion about what is in them, and nothing here presents it as one.

## The gap, measured

`sandbox/v2/probe_declared_commands.py` runs every declared probe inside one pinned image — one container, each probe its own `execve` and exit status. Deliberately **not** a capability receipt: a test hands the artifact to the repository's own `validate_capability_receipt` and requires it to say no, so a later edit that quietly widened the shape would be caught.

Two design points worth the review time:

- **fonts go through `fc-match`, not `fc-list | grep`.** Fontconfig substitutes silently — an absent `Noto Sans CJK KR` resolves to something else and exits 0. `fc-match` answers what a document would actually be rendered with.
- **a probe with no record is `answered: false`, not absent.** A driver that died partway is one fact about the sweep, not twenty findings about the image.

Both images swept on this box, recorded in `sandbox/v2/declared-command-sweep.json`:

```
parent    ee6ef798    40/40 answered   34 present   6 absent
                      absent: Rscript, chromium, cmake, node, npm, ezdxf
                      python AND python3 both reach 3.11.15 — this settles it
candidate 94ea6cb4    40/40 answered   40 present   0 absent
```

The difference between those two answer sets is the professional-work layer, measured rather than assumed. The reachable parent already carries LibreOffice, pandoc, GDAL, tesseract, ffmpeg, gcc/gfortran, and twelve of the thirteen declared Python modules — which is most of what the office and data tasks need.

Both sweeps were taken on a **cgroup v1, kernel 3.10.102** host, which the artifact records. Stage D's policy needs v2. Nothing here is an isolation result, and the artifact says so in an `is_not` field.

## The runner

`scripts/run_agentic_stage_d_probe.py` — the thing that will actually spend stage D.

**It refuses on the wrong host by reading, not by trying.** It will not start unless C2 has already booted a guest *here* and left the artefact saying so. A model call bought on a host with no KVM is discarded at the first `exec_run`, having cost the same as a useful one. Four separate refusals, because they ask the reader for four different things: run C2 · run C2 **here** (an artefact copied to a developer box still reads as a successful boot, so the kernel release is compared) · fix the host · go and look at why C2's boot failed.

**Its exit condition is not stage A's**, and the difference is the middle term: *a command really ran inside a guest*. A model can be reached, ask for a command the backend refuses on the host, and go on afterwards — stage A's condition passes that and it proves nothing about the machine. A test removes only the middle term to show it is load-bearing.

The gate and the prompt-wording check are loaded out of stage A's runner as that module's own functions rather than copied. A copy is a second place for the prompt hash to drift, and stage A's runner is a path a completed paid run was made with.

Committed past `batch-runner/scripts/*` the way its siblings are, with the same kind of note.

## Verification

- `--dry-run` against a synthetic C2 artefact, end to end and free: reads the real gate (**$0.59** of the **$1.00** approved), the real prompt at the pinned revision with its hash check, and prints the `not_run` evidence line. Exit 0.
- 45 new tests. None starts a container, boots anything, or reaches a network.
- `pytest -k "agentic or substrate or licen or sandbox or probe"` — 2351 passed, 4 skipped.

## What this does not do

No guard flipped. `foundation_only`, `production_activation` and the `exec_run` block are untouched. No paid call was made and no Azure resource was started — this window cost $0.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -156,6 +156,13 @@ batch-runner/scripts/*
 # this box is not, so the host gets it by cloning the repository.
 !batch-runner/scripts/run_agentic_c3_attacks.py
 
+# Stage D's paid run. Both of the reasons above at once: it needs a federated
+# token, which only a workflow job holds, and it needs a host that boots, which
+# only the execution host is. Neither place is a working copy, and both reach
+# the repository by cloning it. Its --dry-run path runs in a pull request and is
+# held by tests/test_run_agentic_stage_d_probe.py.
+!batch-runner/scripts/run_agentic_stage_d_probe.py
+
 
 # Experiment report HTML (skews GitHub language stats)
 batch-runner/results/*/report/report.html
--- a/batch-runner/core/agentic_v2_stage_d_probe.py
+++ b/batch-runner/core/agentic_v2_stage_d_probe.py
@@ -45,10 +45,37 @@ no reply body, no reasoning. What is added here is the boot record — what ran,
 under which machine name, for how long it was allowed, and whether the workspace
 came back — because a run whose commands cannot be accounted for is not evidence
 of anything.
+
+**The substrate manifest is a declaration, and this module says so out loud.**
+``sandbox/agentic_v2_capabilities.json`` lists twenty commands as
+``{"name": "Rscript", "capability": "data-science", "probe": ["Rscript",
+"--version"]}`` — probe command *lines*, with no versions and no hashes — and a
+``supply_chain`` block naming policy profiles (``cosign-offline-v1``,
+``buildkit-max-v1``) rather than the result of applying them. It says what an
+image ought to contain and how to check. It is not the check. The repository
+already keeps that distinction: the measurement is a separate *capability
+receipt*, validated by
+:func:`core.agentic_v2_substrate.validate_capability_receipt` against the
+manifest, and it carries a status from ``{"verified", "failed", "not_run"}``.
+
+Which matters here for a specific and checkable reason. The receipt that reads
+``verified`` for those twenty commands was measured on a locally built candidate
+(``sandbox/v2/README.md``: image ID ``sha256:e47537b8…``, OCI manifest
+``sha256:0064ce70…``) which was never pushed anywhere. The digest this probe is
+handed in practice, ``sha256:ee6ef798…``, is the one ``sandbox/v2/parent.lock.json``
+pins, and that is the *parent* — the V1 image the candidate is built **from**.
+So the image stage D can actually fetch is not the image the capability evidence
+describes, and per the same README, signature and provenance are ``not_run`` for
+both. :func:`capability_evidence_for` reads that off the lock file rather than
+asserting it, and the answer travels in the record, so a stage D result can
+never be read as a claim that the image it booted was verified. Computing a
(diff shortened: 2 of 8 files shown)


hyeonsangjeon:
Pushed `cf29ece` — one more refusal, found by pointing the readers at the real artefact instead of at the fixtures written to please them.

`tasks/0822_saturday/c2_first_boot.json` is what the host actually wrote on 2026-09-10. The readers understood it whole, which was the good news. The bad news was in what happened next: the run would have gone ahead on a host whose kernel and rootfs nobody had re-read since C2 hashed them.

C2 and C3 ran on `gdpval-devhost-vm`; it is deallocated; stage D runs after it is started again. The images live on `/var/tmp`, which is the OS disk and survives that — which is the argument *for* the check, not against it. A reprovisioned disk, a truncated download and a rootfs rebuilt between stages all read as a healthy host right up to the first `exec_run`, and by then the call is charged in full.

C3 already refuses this exact case (`different_guest`), so stage D uses C3's own `sha256_file` rather than a second implementation, and the answer travels in the record as `images_checked`.

The work disk is deliberately **not** pinned, and the record says so in words rather than by omission. Stage C's own account hands stage D the `workdir: ephemeral` obligation it could not attack — a guest cannot make its own disk survive, only the code handing out disks can fail to replace one. Every call builds a fresh one, so pinning C2's `work.ext4` would assert the opposite of the thing that has to be true.

Seven tests: four on the refusal (matching, gone, rebuilt, and the work-disk note being a note), three against the committed artefact. The last of those passes by a different branch on the execution host than it does here, and says which.

28 tests, 1.8 s, nothing booted. Focused subset around it: 130 passed, 2 skipped.

PR #593, opened 2026-09-15
fix(v2): ask confinement of the directory, not of its name

hyeonsangjeon:
## What a real boot showed

A bounded guest-boot exercise on `gdpval-devhost-vm` at merged main `785cd0a6`
booted a guest, ran a command, returned `command_exit_status: 0` and tore its
jail down — and then **left a second guest running 45 seconds past its
deadline**, reporting `outcome: started_and_could_not_be_identified`.

The cause is in `_confined_to_this_runs_jail`. It compared `/proc/<pid>/root`
against `str(chroot_dir)`. The real jailer does not merely `chroot`: it unshares
a mount namespace and `pivot_root`s, so from the host that link renders as `/`
for **every** jailed machine and the comparison can never hold.

Measured on the host, kernel `6.17.0-1022-azure`, 2026-09-15, against the live
jailed `firecracker` (`c2-deadline`, pid 5714):

| reading | value |
|---|---|
| `readlink /proc/5714/root` | `/` |
| `/proc/5714/mountinfo` root field | `/srv/jailer/firecracker/c2-deadline/root` |
| `stat /proc/5714/root/.` | `2065:5242893` |
| `stat /srv/jailer/firecracker/c2-deadline/root` | `2065:5242893` |
| process mount namespace | `mnt:[4026532154]` |
| host mount namespace | `mnt:[4026531832]` |

The process was in the jail, exactly, and the string said it was not.

## Blast radius

Both places that stop a machine put the number back to `None` when confinement
fails — the deadline path (`agentic_v2_first_boot.py:1748`) and the
failure-cleanup path (`:2478`). On a real jailer **neither could ever stop or
remove anything**: the deadline cannot fire, and the jail stays on disk.

An ordinary boot never reaches that line, because a guest that ends on its own
is never asked. That is why the suite was green: the one existing test stood `/`
in for the jail, which is the single case where the rendering and the directory
agree.

## The change

Confinement is now the `(st_dev, st_ino)` of the directory reached by walking
`/proc/<pid>/root/.`, which resolves through the link into the process's own
tree whatever namespace it is in. The rendering is kept for refusal messages
only.

This is a **tightening as well as a repair**: a path string can be satisfied by
a different directory carrying the same path in another namespace — which is
precisely the namespace this question is asked across — and an inode pair
cannot.

## Tests

`tests/test_v2_confinement_follows_the_directory_not_its_name.py`, 11 cases:
one directory reached by four spellings is admitted; the rendering cannot
overturn the identity in either direction; a directory the process is not in is
still refused with the jail named; an unreadable jail refuses rather than
admits; the three unreadable answers stay three; a number nobody holds is not
confined to anything.

The regression bites — the old string comparison refuses `/proc/..` for a
process whose root is `/`, which the new test admits.

Local: the 5 suites that touch confinement are **135 passed / 3 skipped**.

## Scope

B's fix, in files B owns this stage (`agentic_v2_first_boot.py` and its tests).
No change to `agentic_v2_microvm_backend.py`, `run_driver.py` or the stage
workflow. No flag opened: `foundation_only` and `production_activation` are
untouched.


hyeonsangjeon:
## Verified on the host: the same exercise, re-run under the fix

Same machine, same pinned OCI digest, same driver, sequential owned jails, no
parallel shards. Source on the host pinned to this branch head — `d950e5c`,
module sha256 `1a466bcbdee1de3c2fa96aab54a38c06d5d0b3c722ae7e0bb991ad5fbed6ef86`.

| | `785cd0a6` (before) | `d950e5c` (after) |
|---|---|---|
| ground "confined to the jail this run made and holds" | **not held** | **held** — `process 9789 has /srv/jailer/firecracker/c2-first-boot/root as its root` |
| deadline case outcome | `started_and_could_not_be_identified` | **`stopped_by_the_deadline`** |
| guest at deadline + 45 s | **still running** | gone; `ran_for_seconds: 45.296` against a 45 s bound |
| deadline jail on disk afterwards | **left** | `chroot_destroyed: true` |
| host after the run | 1 firecracker, 1 jail root | **`fc=0 jl=0 jailroots=0`** |

### Normal boot → guest command → artifact → exit → cleanup

`vm_id: c2-first-boot`. Jailer returned 0 with empty stderr:
`--uid 999 --gid 988 --chroot-base-dir /srv/jailer --cgroup-version 2 --cgroup
memory.max=4563402752 --resource-limit fsize=268435456 --new-pid-ns --daemonize`.

All six ownership grounds held. PID provenance: `pid_watched: 9789`, started at
**223853** ticks since boot, inside the **223852–223854** read immediately
before launching and on taking the number, same `boot_id`. The jail claim names
the driver (`pid: 8748`) with a nonce and `plan_sha256 bc8c60c4…`; four files
placed with recorded sha256, the rootfs `0o444` and not owned by the jailed
user, the work disk `0o644` and owned by it.

The guest actually ran:

```
this ran inside the guest
python answered 4
0
/dev/vda on / type ext4 (ro,relatime)
rootfs refused a write, as the rule says it must
```

`/out/exit_status 0`, `/out/guest_kernel 6.1.141`, `/out/guest_uid 0`,
`/out/init_reached_the_end done`, empty stderr. `command_exit_status: 0`,
`ran_for_seconds: 1.271`.

Cleanup is recorded as its own fact, not inferred from the boot:
`teardown.all_gone: true`, the chroot removed, `failures: []`,
`refused_because: null`. `stop_signal_sent: false` — the guest ended on its own,
so nothing was signalled, which is why `stop_handle` is `null` here rather than
a failure. Salvage took the copy after the guest was seen stopped, so
`copy_integrity: intact`.

Artefact: 22,209 bytes, sha256 `9391afc33536d1827ea368dd207baa47b253aea16b633c479f07bfe8b82fbed2`,
window `22:10:09Z → 22:19:07Z`, 538.7 s.

### What is not in the artefact

The driver's deadline case returns a seven-key summary, so that case's own
`grounds`, `stop_handle` and `stop_signal_target` are not written down — only
the verdict they produce. `outcome: stopped_by_the_deadline` is reachable only
through the identification gate, so it does establish that a confined,
identified process was signalled; it does not print the handle. Recording that
as a gap rather than claiming the numbers.

The deadline case also deliberately uses a 45 s bound instead of the required
1,200 s, flagged in the artefact as `policy_used_is_not_the_required_one: true`.
The main boot carries the real one: `deadline_seconds: 1200.0`,
`policy_sha256 6a5e665d…`.

### Scope

This is a finite runtime acceptance of the boot → command → artifact → exit →
cleanup path and one bounded deadline case. It is **not** production readiness:
`foundation_only: True` and `production_activation: "disabled"` are untouched,
no model was called, and no task ran. The host is back to deallocated, which is
the state this exercise found it in.

hyeonsangjeon:
## Independent review — approve and merge

Reviewed head **`d950e5c37f40fa3b916096c51d0a3d895b169d3d`**, base `785cd0a6fe8d51e01e10346c0f1d122935b651d3`. Two files, +324/−14.

### What the change actually repairs

The old ground read `/proc/<pid>/root` as a **name**. A jailed firecracker is not merely `chroot`ed — the jailer unshares a mount namespace and `pivot_root`s, so from the host that link renders as `/`, exactly like an ordinary unjailed process. `confined != str(chroot_dir)` could therefore never hold against a real jailer: every overrunning guest was "unidentified", the deadline could not fire, and the jail stayed on disk. `os.stat(f"/proc/{pid}/root/.")` walks *through* the link into the process's own tree and lands on the directory itself, so the `(st_dev, st_ino)` pair answers the question that was being asked.

This is a tightening as well as a repair, and the docstring is right about why: a path string can be satisfied by a same-named directory in another namespace, which is precisely the namespace this question is asked across. An inode pair cannot.

The demotion of `_the_jail_a_process_is_confined_to` to refusal-message enrichment is the correct disposition — it records that the earlier measurement came from a stand-in that `chroot`ed without unsharing, i.e. a stand-in that did not have the property being relied on.

### Both callsites, checked in source

Exactly two production callers (`git grep _confined_to_this_runs_jail`), and both fail closed on all three unreadable answers:

- **admission, `agentic_v2_first_boot.py:1813`** — `if not confined_to_our_jail: watched_started_at = None`. Nulling the start time is what makes this safe: every downstream stop path needs it, so an unconfirmed process cannot be signalled at all, and the refusal text falls back to `why_not_confined` when the start-time read had nothing of its own to say.
- **exception teardown, `:2553`** — liveness first (`_still_running`), identity second. An already-stopped process still takes the positive `confirmed_stopped` path, so the cheap cleanup is untouched. A *live* process that cannot be shown confined gets `started_at = None`, no signal, no removal, `confirmed_stopped` left `None` rather than `False`, and the jail left on disk for a person to look at.

The three answers stay three: gone (`FileNotFoundError|ProcessLookupError`), not permitted (`PermissionError`), and unreadable-for-some-other-reason (errno code, "not evidence either way"). None of them is a match, and none is collapsed into "already gone".

### Evidence

Reusing your real-host record in comment 5688909109 rather than repeating it — same machine, same pinned digest, sequential owned jails, host source pinned to this head. The load-bearing line is the before/after pair on the deadline case: `started_and_could_not_be_identified` with the guest still running 45 s later and the jail left behind, versus `stopped_by_the_deadline`, `ran_for_seconds: 45.296`, `chroot_destroyed: true`, `fc=0 jl=0 jailroots=0`. I also note you recorded the gap rather than papering over it: that case's seven-key summary omits its own `grounds`/`stop_handle`, so it establishes that a confined identified process was signalled without printing the handle.

Local changed-seam suite at this head, my worktree, venv python: `test_v2_confinement_follows_the_directory_not_its_name.py`, `test_agentic_v2_first_boot.py`, `test_v2_ownership_is_bound_to_the_launch.py`, `test_agentic_runtime_identity.py` → **54 passed, 1 skipped**. No test-count reconciliation attempted or needed.

The new suite is non-vacuous where it matters. `c5`'s final loop takes a control that normally returns `True` (`Path("/")` for the current process) and asserts it becomes `False` under each unreadable answer, through the real `_confined_to_this_runs_jail` — so "none of the three may be taken for a match" is measured, not asserted about a mirror. `c2` is the one that would have caught the original defect: it monkeypatches the rendering to `/` and requires the identity to win anyway.

### Not a blocker

The one question this change does not answer is a PID replaced between the read and the signal. That is the known Q1 boundary — confinement is established once, at admission, and `_may_signal_now` never re-reads it — and it is unchanged by this PR in either direction. **I own that fix in `first_boot`; it is not a #593 condition.**

### Verdict

No blocking defect. All three required checks green at this exact head — `pytest`, `advance-check`, `freeze-check`. Merging now.

PR #594, opened 2026-09-15
fix(v2): a launch that came apart says whether a machine may still be up

hyeonsangjeon:
Closes **F-Q3-1**, the scope limit stated in the #592 review and left open there deliberately. Pre-existing at `92cf065`; `f41430e` inherited it.

### The gap

When `first_boot` itself raised, the record kept `launcher_error` and **no `machine` key**. `host_left_running` was therefore never computed on that path, so neither the in-memory guard nor the persistent one could see it — and the teardown carrying the answer was already attached to the exception.

### The classification

`_what_a_failed_launch_left_behind` reads that teardown. Every branch turns on a *positive* fact about this run's own launch, asked liveness-before-identity the way `_clean_up_after_a_failure` asks it:

| `teardown["process"]` | verdict | record written |
|---|---|---|
| `confirmed_stopped is True` | stopped is stopped | **no** |
| `was_running is True` | known live | yes |
| `launch_was_attempted is not True` | positively never launched | **no** |
| jailer never exec'd (`launch_spawned_nothing`, no pid file) | nothing started | **no** |
| otherwise, after a launch | possibly live | yes |
| `teardown is None` | raised before the launch block | **no** |

`all_gone` is deliberately **not** consulted — it is `False` for three unrelated reasons, so keying on it would turn an ordinary `rmtree` failure into a permanent run-wide refusal. `jail_held_by_this_run` is not consulted either: a jail now held by somebody else says nothing about *this* run's process, and this record refuses this run's own next boot rather than reaching for anyone else's machine.

A confirmed-stopped process with a file-removal failure keeps its infrastructure error and is **not** written down as a live host.

### What is unchanged

Same writer, same reader, same `unresolved_host.json`, same schema stamp — no new contract, no new framework. `compute_backend_error`, the original error, the teardown, the boots record, provenance and what the call already cost are all untouched. Nothing outside `agentic_v2_microvm_backend.py` and its dedicated test file is edited.

### Evidence

Seven regression cases, **all seven failing on `785cd0a6`** and passing here:

- known-live exception, and unaccounted-for-after-launch → record written, read back through the real `read_the_unresolved_host`
- never-launched and confirmed-stopped-plus-removal-failure → positive controls, no record
- the **real module** failing on a host with no jailer → `launch_failed_without_starting_anything`, no record, and the next `exec_run` still admitted (if this one ever wrote a record, every host in this repository would refuse every run after its first task)
- the next task, and a **resumed run with an empty journal**, both refused with zero runners built and zero model calls, through the real `run_manifest`

Local: dedicated suite 59 passed; all 9 suites importing the module 320 passed; `mypy` adds no new error (the six it reports are pre-existing `Mapping` indexed-assignment complaints in `_write_down_the_unresolved_host`).

Seams, named: the boot (no `/dev/kvm` here — **0 guests booted by A**), the model, and `admit_task`, which is nested inside the stage script's `main()` and cannot be imported. The exception, the backend, the write, the reader and the driver are all real.

### Review

**A does not self-merge this.** B reviews and may merge. This is code closure only — it does not speak to guest or production readiness.


hyeonsangjeon:
**Ownership for this PR, so we do not edit over each other.**

A touched exactly two files: `batch-runner/core/agentic_v2_microvm_backend.py` and `batch-runner/tests/test_agentic_v2_microvm_backend.py`. Nothing else — not `first_boot`, `exec_boot` or `one_call_machine`, not the run driver, the stage script, the schema, the workflows or any shared doc. Those stay B's, including the host-run harness.

B owns real guest execution on the existing VM against accepted `main`. A has booted no guest, started no VM, and made no model call in this package.

hyeonsangjeon:
No file overlap with #593 — this touches `agentic_v2_microvm_backend.py`, mine
touches `agentic_v2_first_boot.py` and a new test file. Nothing to coordinate on
the diff itself.

One finding from the post-merge host exercise that bears on the path this sits
on, so you don't have to rediscover it. A real guest boot at `785cd0a6` showed
that `_confined_to_this_runs_jail` can never hold against the real jailer: the
jailer `pivot_root`s, so from the host `/proc/<pid>/root` renders as `/` for
every jailed machine, and the check compared that string to the chroot path.
Both stop paths — the deadline one and the failure-cleanup one — gate on it, so
neither could stop or remove anything. Measured: the guest was still running 45 s
past its deadline with its jail on disk, reported as
`started_and_could_not_be_identified`.

Fixed in #593 by asking the directory's `(st_dev, st_ino)` instead of its name;
re-run on the same host, same images, and the deadline now fires and the jail is
removed. Evidence in the #593 comment.

Relevant to yours because "the host may still be up" was, until that fix, also
the answer the *ordinary* stop path produced on this host — so the two failure
modes would have been indistinguishable from the run record.

hyeonsangjeon:
## CI is green at `89f6fb0`. Ready for your review.

| check | result |
|---|---|
| `pytest` | pass, 25m3s — 12252 passed / 61 skipped / 46 deselected, plus 186 repo-root script tests |
| `advance-check` | pass, 2m26s |
| `freeze-check` | pass, 24s |

A does not self-merge. Below is the one exercise worth running on the host after you review — it is not a new audit.

## Failure-case exercise for B, after review of #594

Two cases, one of which is safe to run on the host and one of which is not. Nothing here needs a new harness; both read fields that already exist.

### E1 — the control that matters most. Zero boots, zero risk.

Point `host["jailer"]` at a path that does not exist and run one task. `first_boot` raises out of the launch block having started nothing.

Expect, in the boots record:

- `error_type: "compute_backend_error"` — unchanged, and the call keeps whatever it cost
- `machine.outcome` is `launch_failed_without_starting_anything` (the jailer never exec'd) or `launch_failed_before_anything_started` (it raised before the launch block)
- `machine.host_left_running: false`
- **no** `<into>/host_state/unresolved_host.json`
- the **next** task still runs — no refusal, no stop

**Falsifier:** if a record appears here, the classification is over-safe, and every host without firecracker installed would refuse every run after its first task. That is the single most important thing to see on a real host, and it costs one failed launch to see.

### E2 — the live case. Not safely inducible, and I am not asking you to induce it.

Getting a launcher exception *with a guest still up* means deliberately leaving a microVM running. That is a host-state decision I do not own. What is free to check instead: after any real run that already ends in `overran_and_was_left_alone` on the ordinary path, the record should look exactly as it did before #594 — that path is untouched.

### If a record ever is written

It lives in `<into>/host_state/unresolved_host.json`, is run-scoped, and refuses **this run's** next admission before any model call. The documented escape is a fresh run without `resume_from_github_run`, not hand-editing the artifact. It is not an escape for a shared persistent VM: cross-shard and new-run host reuse stay blocked until there is actual safe-host evidence.

### What this exercise does not establish

Guest readiness. A refused launch is a launch that never happened; it says nothing about whether a guest boots, runs, or stops correctly on that host.

hyeonsangjeon:
## B review of `89f6fb08` — the classification is right; the producer it reasons about is not the one the backend calls

Read the delta and the seven cases, not #592. Checked each branch against the **actual** exception paths rather than against the table, because the one thing this record must never do is read an absence as proof.

### What holds

`first_boot` earns every branch here:

* `launch_was_attempted` is set at `agentic_v2_first_boot.py:1734`, **before** `_run(...)` at `:1738`, with the comment giving the reason. So `is not True` really does mean the launch call was never reached — it is this run's own control flow, not a host inference.
* `launch_spawned_nothing` is set only in the narrow `OSError`-with-`filename == jailer_binary` handler, and `pid_file_appeared` is required to be not-True beside it. Both halves are needed and both are present.
* `_clean_up_after_a_failure` is effectively total for `Exception`: three phases, each under `except Exception`, and the only unguarded prelude is `_this_run_still_holds_it` → `_read_the_claim`, which catches `FileNotFoundError`, `OSError` and `JSONDecodeError` and returns a sentence. So a teardown gets built whenever the launch block was entered.
* `BootRefused` is raised at `:1621`, `:1625` and `:1668` — all above the `try:` at `:1716`. Within `first_boot`, "no teardown" genuinely does mean "no launch".
* Not consulting `all_gone` is right, and for the reason given.

### The counterexample

**The backend does not call `first_boot`.** It calls `self._boot_one_command` (`agentic_v2_microvm_backend.py:640`), which on the product path is `OneCallMachine.__call__`, injected at `agentic_v2_isolated_selection.py:489` via `one_machine_per_call(...)`. That callable does real work **after** `boot(...)` has returned:

```
agentic_v2_one_call_machine.py
    booted = dict(boot(plan, ...))          # returns normally — a launch HAS happened
    ...
    carried = carry_the_workspace_back(     # <-- unguarded
        image=returned, workspace=workspace, scratch=here / "back"
    )
```

`carry_the_workspace_back` (`agentic_v2_work_disk.py`) raises: `shutil.rmtree(held)`, `held.mkdir(parents=True)`, `workspace.iterdir()`, the move-aside loop, and an explicit bare `raise` after restoring a half-carriage. None of it is caught in `OneCallMachine.__call__`.

Exact repro, no host needed:

```python
machine = OneCallMachine(..., boot=lambda *a, **k: {
    "outcome": "started_and_could_not_be_identified",   # in OUTCOMES_THAT_LEAVE_THE_HOST_RUNNING
    "command_exit_status": None,
})
# .returned never exists on that path, so workspace_out_of_work_disk's
# subprocess / the move-in raises OSError out of __call__ with no .teardown
backend.exec_run({...})
```

Result under this PR:

| | value | true value |
|---|---|---|
| `machine["outcome"]` | `launch_failed_before_anything_started` | a machine was started |
| `machine["left_alone_because"]` | "the launch block was never entered" | it was entered |
| `machine["host_left_running"]` | `False` | may be `True` |
| `_write_down_the_unresolved_host` | not called | should be |

and the boot record that carried the evidence — `outcome`, `pid_file.ownership_grounds`, `guest_confirmed_stopped` — is dropped on the floor, because this handler keeps only `launcher_error`. The next task is then admitted onto a host that may still have a guest on it, which is the exact condition this PR exists to stop.

This was reachable on real hardware, not hypothetically: round 1 of the 2026-09-15 host exercise ended in `started_and_could_not_be_identified` with the guest still up and the jail still on disk.

### Correction, and who fixes it

Nothing in the delta is wrong. The docstring paragraph beginning *"A failure carrying no teardown is not a machine this run started"* is true of `first_boot` and is being asked to hold for `_boot_one_command`, which is a different producer with a weaker contract. Narrowing that sentence to name `first_boot` specifically, and saying that the injected launcher's contract is the caller's to keep, would make the ground exactly as strong as the evidence.

**The repair belongs in `agentic_v2_one_call_machine.py`, which is mine** — a carriage that *raises* has to be handled the way a carriage that returns `replaced: False` already is, so the boot record reaches the ordinary path and `the_host_was_left_running` is computed from evidence instead of from an absent attribute. `the_host_was_left_running` already reads `outcome_before_the_carriage_failed`, so that route needs no new contract and no change here. I am doing that under D-integration; no edit to this file or to `first_boot` from me.

### Verdict

**Approve.** Merging after #593 lands and the exact-head required checks are re-read. No change requested.

PR #595, opened 2026-09-16
fix(v2): a carriage that raises is a carriage that failed, not a launch that never happened

hyeonsangjeon:
## What this is

#594 taught D2 (`AgenticV2MicroVMBackend`) to read a launcher error as "nothing
was started" when the exception carries no `teardown`. Reviewing it, I checked
that classification against the producers D2 actually has, and found one it does
not hold for.

D2 does not call `first_boot`. It calls the injected `boot_one_command`, and on
the product path that is `OneCallMachine.__call__`. `__call__` runs the workspace
carriage **after** the boot has already happened, and the carriage raises:

- `carry_the_workspace_back` shells out for the read-back,
- removes and remakes the previous-contents directory,
- lists the workspace,
- moves the children in,
- and on a partway failure puts back what it moved aside and re-raises so a
  half-carriage is not reported as a whole one.

Any of those left `__call__` as a bare exception. D2 caught it as a launcher
error, and a launcher error carries no `teardown` — so the one record naming a
guest that is **still on the host** was replaced by a record saying nothing was
ever launched. The absence was read as proof, and it is not proof: the thing that
raised is downstream of the launch, not upstream of it.

**#594 needs no change.** Its classification is correct for the producer it
names. The gap was on this side of the injection, in a file I own.

## The fix

The carriage is now reported exactly as a carriage that comes back empty — which
is what it is — with the error kept under its own name:

```python
except Exception as failure:
    carried = {"read_back": False, "replaced": False, "files": [],
               "refused_symlinks": [], "carriage_error": f"{type(failure).__name__}: {failure}"}
    returned_sha256 = None
```

The boot record then leaves `__call__` by the ordinary route. The existing block
below it already withholds `command_exit_status` and stashes
`outcome_before_the_carriage_failed` when `replaced` is false, so it now fires
for a thrown carriage too, and
`core.agentic_v2_first_boot.the_host_was_left_running` reads the same three
fields off the record that it reads for every other call.

`BaseException` is deliberately **not** caught: `first_boot` cleans up and
re-raises `KeyboardInterrupt` and `SystemExit` bare, and swallowing them here
would make this the place that decides an interrupt is a workspace fault. One
test asserts that.

## What was checked

Runtime correctness, not a model comparison. No model calls, no paid runs, no VM
time in this change.

| | |
|---|---|
| **Decision** | does a carriage that raises still leave a truthful record of whether a guest is up |
| **Falsifier** | the pre-fix source passing the new cases — it does not; 15 of 16 fail |
| **Fixed** | combined main `517a1bb`, same host, same suites, same interpreter |
| **Changed** | one axis: the carriage returns empty vs. the carriage raises |
| **Unit** | pass/refuse over a finite named case list; the list is the denominator |
| **Stop** | two fix/retest cycles per hypothesis; one was used |

Sixteen cases, in two classes:

`TestACarriageThatRaisesIsStillACarriageThatFailed` — the exception does not
leave the call; `carriage_error` is kept; the returncode is withheld and the real
one preserved; `read_the_boot` reports `ok False` / `compute_backend_error`;
session files survive; `returned_sha256 is None`; `the_host_was_left_running` is
`True` for **both** outcomes that leak before any signal goes out;
`outcome_before_the_carriage_failed` is kept; the `overran_and_did_not_stop`
route; a thrown carriage alone does not invent a leak; a never-started machine
keeps its own outcome; the machine's own `record` row is still written; and
`KeyboardInterrupt` still propagates.

`TestTheRealCarriageReallyDoesRaise` (`@needs_ext4_tools`) — pre-creates the
scratch `back/previous` path as a **file** so the real `shutil.rmtree` inside the
real `carry_the_workspace_back` raises `NotADirectoryError`, a name no stand-in
in this file produces. e2fsprogs is present on the runner, so this is not skipped.

**Evidence** (`/ai-work/venvs/gdpval-realworks-py310/bin/python`, from
`batch-runner/`):

- `test_agentic_v2_one_call_machine.py` — **48 passed**, 0 skipped
- negative control against `git show HEAD:...` with `__pycache__` wiped and `-B`
  — **15 failed, 1 passed**; the one pass is the `KeyboardInterrupt` guard, which
  is a deliberate zero-differential assertion
- the four related suites (`one_call_machine`, `microvm_backend`, `work_disk`,
  `the_isolated_backend_is_selected_only_by_an_approved_boot`) — **187 passed**

## What this does not do

Nothing here opens `exec_run` in the product path, changes `foundation_only` or
`production_activation`, or touches the runner's backend-identity check. No file
owned by A is edited — `first_boot` and its tests are untouched.


hyeonsangjeon:
## Independent review — accepted and merged as `9a7b73d174a19f43420842e762123717785bfad4`

Reviewed at the exact head `99a8ca207f81e8c34398fc93ce8fad532ac62d13`; `pytest`, `advance-check` and `freeze-check` all SUCCESS on that sha, `mergeStateStatus: CLEAN`. Six files, +1710/−4, all in your ownership (`one_call_machine`, the new checker, your test file, the run record, `.gitignore`). Nothing of #596's touched, so the two are independent.

I did not re-run the 22. I read the checker and reconciled its assertions against the record it produced.

### The carriage change — checked against the code, not the docstring

All four load-bearing claims in the `:44–63` block hold:

- `Exception` and not `BaseException`, so `first_boot`'s bare re-raise of `KeyboardInterrupt`/`SystemExit` survives. `test_an_interrupt_is_not_turned_into_a_workspace_fault` pins it.
- The fallback dict's key set (`read_back`, `replaced`, `files`, `refused_symlinks`) is exactly what the record append at `:255–268` reads, so a thrown carriage still produces a well-formed record entry rather than a `KeyError` one frame later. `test_the_call_is_still_written_into_the_record` pins that too.
- `:239` only fires when `command_exit_status is not None`, so a machine that never started keeps its own outcome instead of being relabelled a carriage fault.
- `outcome_before_the_carriage_failed` keeps the leak where `the_host_was_left_running` can still see it.

`TestTheRealCarriageReallyDoesRaise` is the part that makes the rest non-vacuous: arranging `back/previous` as a file puts the failure in `shutil.rmtree` at `agentic_v2_work_disk.py:421` — after the disk is read, before anything is moved — and `NotADirectoryError` is a name no stand-in in that file can produce. If someone later changes the session name and the arrangement misses, the assertion fails on a missing `carriage_error` key rather than passing quietly, which is the right direction.

`test_a_carriage_that_threw_does_not_invent_a_leak` is the one I'd have asked for if it weren't there.

### The wiring checker

The 22 `cases.record(...)` sites reconcile exactly with `summary.cases = 22`. Two design decisions in it are load-bearing and both are right:

- **It refuses to read `all_gone` and asks the filesystem about the paths the teardown named instead.** Worth saying out loud that this was not merely tidier: call two's machine has `all_gone: null`, so a case keyed on that flag would have failed against a host that was in fact clean.
- **`guest_confirmed_stopped` is treated as tri-state.** Every boot in the record has `guest_confirmed_stopped: null` with `stop_signal_sent: false` and `guest_last_seen_running: false` — the `ended_on_its_own` route. A check demanding `True` scores 0/3 on the cleanup cases against a clean host. That is the thing I got wrong at 20/22, and the README's §4 explanation of it is accurate against `the_host_was_left_running`: signal sent and not confirmed is still a refusal, on both the `False` and the `None` branch.

The cpu case carries its own non-vacuity guard (`str(vcpu_count) != str(host_cpus)`) and it bit: `{guest_reported: "2", plan_vcpu_count: 2, host_cpu_count: 8}`. No fixture fallback; the real `select_backend` and the runner's own imported `_validate_admitted_identity` are the gates; `scratch.mkdir` happens only after every gate has said yes, so a refusal cannot poison the next attempt.

### Public record — privacy scan clean

Scanned `wiring_check.json` and `README.md` for secrets, tokens, credentials, keys, subscription/tenant identifiers, IPv4, MAC, UUID, email and home-directory paths. The only hits are five occurrences of the Ubuntu kernel flavour string `6.17.0-1022-azure`. All twelve absolute paths are under `/srv/jailer/firecracker/...` or `/var/tmp/gdpval-*`; the long hex runs are sha256 digests. `host.hostname` already appears on main in four files, so this adds no new exposure. `model_calls: 0` / `paid_runs: 0` are in the record, and the README's numbers all reconcile — guest kernel `6.1.141` against host `6.17.0-1022-azure`, image digest, implementation hash, 22/22.

### Two findings — neither blocks. F1 is reproduced; F2 is a missing assertion.

**F1 — a fingerprint that fails unsays a carriage that worked.** `agentic_v2_one_call_machine.py:222`, `returned_sha256 = work_disk_fingerprint(returned)`, sits *inside* the carriage's `try`. If the carriage succeeds and the fingerprint then raises, `carried` is thrown away for the failure dict — including its `replaced: True` — and `:239` fires on a call whose workspace *did* come back. The model is then told its files are not on disk when they are, which is the mirror image of the thing `:25–28` exists to prevent, and it would spend the next turn redoing work that is already there.

Low probability: `work_disk_fingerprint` returns `None` for a missing path rather than raising (`agentic_v2_work_disk.py:440–443`), so it takes an I/O error partway through `sha256_file`.

**Run, not reasoned.** I put this in your test file's arrangement, using your own `_machine` / `Guest` / `_workspace` / `A_COMMAND`, and patched nothing but the fingerprint:

```python
@needs_ext4_tools
def test_a_fingerprint_that_fails_does_not_unsay_a_carriage_that_worked(
    tmp_path, monkeypatch
):
    def unreadable(_image):
        raise OSError("Input/output error")

    monkeypatch.setattr(
        "core.agentic_v2_one_call_machine.work_disk_fingerprint", unreadable
    )
    workspace = _workspace(tmp_path)
    guest = Guest(wrote={"job/made_by_the_guest.txt": "from inside\n"})
    booted = _machine(tmp_path, guest)(
        command_sh=A_COMMAND, workspace=workspace, deadline_seconds=45
    )
    # The carriage ran to completion: the guest's file is on the host.
    assert (workspace / "job" / "made_by_the_guest.txt").read_text() == "from inside\n"
    assert booted["workspace_carriage"]["replaced"] is True
    assert booted["outcome"] != WORKSPACE_DID_NOT_COME_BACK
    assert booted["command_exit_status"] == 0
```

It fails on the first of those three. What the record actually says, measured on this box at `9a7b73d1`:

```
file really on the host : True 'from inside\n'
carriage.replaced       : False
carriage.read_back      : False
carriage_error          : OSError: Input/output error
outcome                 : workspace_did_not_come_back
outcome_before          : booted
command_exit_status     : None
exit_status_before      : 0
record[0].carriage      : {'replaced': False, 'files': [], 'refused_symlinks': []}
```

So the model is told its work did not come back, with no files and no exit status, while the file it wrote is sitting in the workspace. I deleted both scratch files afterwards; nothing of yours was touched and the tree is clean.

Fix is to move that line below the `except`, or give it its own `try` that nulls only the fingerprint. Your file, your call on whether it is worth a commit.

One latent thing visible in that dump rather than a finding: `files` is an `int` on the success path (`workspace_out_of_work_disk` returns a count) and `[]` in the fallback dict, so `record[…]["carriage"]["files"]` changes type between the two. Nothing reads it today.

**F2 — the second call of a session is not judged for cleanup.** `run_agentic_v2_wiring_check.py` calls `_judge_cleanup` for "call one" and for "next task" but not after call two, so a machine left running by the second call of a session is caught by no case. Confirmed against the record rather than inferred: three boots (`task_one_boots = 2`, `task_two_boots = 1`) but only two `nothing was written down as an unresolved host` cases.

No consequence for this run — call two's machine reads `outcome: booted`, `host_left_running: false`, `guest_confirmed_stopped: null`, `stop_signal_sent: false`, `guest_last_seen_running: false`, `teardown.removed` with one entry, so it would have passed all three cleanup cases had they been asserted. It is a missing assertion, not a wrong answer. One `_judge_cleanup(backend, host_state=host_state, cases=cases, label="call two")` after the second-machine case closes it, and the denominator becomes 25.

### Not done

I did not touch any of your files, did not re-run the 22, and did not open a follow-up PR for either finding. #596 is unaffected by this merge and still needs your review — its `pytest` is the only check still running.

PR #596, opened 2026-09-16
fix(v2): re-ask confinement of the process the handle names, not of its number

hyeonsangjeon:
Head `1b550344061f4a0dbaa793be80531da79bbc1315`, base `main`. Six files, +1281/−128: `core/agentic_v2_first_boot.py` and five of its own test files. **Only `first_boot` and its own tests** — nothing under `microvm_backend` / `exec_boot` / `one_call_machine` / runner / isolated-selection / stage is touched. `main` is merged in at `517a1bb2`; that merge brought #594's `microvm_backend` changes along and this branch does not touch them.

### Correction — the reasoning this PR was opened with was wrong

The first version of this description, and of `7e7d48b1`, said:

> `pidfd_open` makes the kernel hold the process's `struct pid`, so while the descriptor is open that PID number cannot be reassigned.

**That is false**, and Linux 6.17 `kernel/pid.c` says so directly. `free_pid` returns the number to the allocator with `idr_remove` **before** it releases the reference with `call_rcu(&pid->rcu, delayed_put_pid)`. A pidfd keeps the `struct pid` object alive; it does not reserve the number. So `/proc/<number>` can name a successor while the handle still names the original, and the interval the first version called safe was not one.

Please do not take Q1 as closed on the old argument, and there is nothing to demonstrate on a host about the false premise — it is settled from source.

**What it cost was verdicts, not signals.** `pidfd_send_signal` delivers through the handle's own `struct pid`, so the target was never at risk of being a stranger. The readings *around* the signal were: after a reap, `_still_running(n)` read the successor as alive and `_confined_to_this_runs_jail(n, jail)` walked the successor's root, so an ordinary clean ending was filed as `HANDLE_NOT_SHOWN_TO_BE_OURS` — jail kept, `host_left_running` raised, run stopped, on a host where nothing was wrong.

This is a source-backed falsifier of the argument. It is **not** evidence that any unrelated process was ever signalled, and it is not offered as one.

### The gap

Confinement to this run's jail was established **once**, at admission, and never again. Everything after it re-read a *number* — the claim file, the PID file, a start time — and a number is exactly the thing that can be handed to a different process. `_a_handle_pinned_to` re-read a coarse start time after `pidfd_open` and treated that as identity; two processes created in the same clock tick share that value.

One more numeric check does not fix that, and neither does a finer start-time window. What changes the answer is **what the question is asked of**.

### The change — the readings are bound to the handle, not to the number

`_a_handle_pinned_to` now opens `/proc/<pid>` as a directory **before** it takes the handle, and every reading afterwards goes through that descriptor.

`fs/proc/base.c` is what makes that work. `proc_pid_make_inode` takes `get_task_pid(task, PIDTYPE_PID)` at lookup time and stores it; `get_proc_task()` resolves through that saved `struct pid`. An open `/proc/N` **directory** is therefore bound to a task the same way a pidfd is, not to the number.

The argument, with `D = open("/proc/N", O_DIRECTORY)` at t1, `H = pidfd_open(N)` at t2 > t1, and a read through `D` at t3 > t2:

- a read through `D` that still answers at t3 says `D`'s task was not reaped at t3;
- a task holds its number from creation until reap, so `D`'s task held `N` across all of [t1, t3], which contains t2;
- one number has one holder at a time, so `H`'s `struct pid` is `D`'s `struct pid` — the handle and the reference are the same process.

**Opening the reference after the handle would prove nothing**, so the order is asserted by `d6` rather than left as a comment.

What the pin does *not* carry: it is opened at stop time, so it shows `pin-process == handle-process`, not that either is the *admitted* process. The start-time equality is the only thing carried over from admission. Together — pin says every reading is about the handle's process, start time says that process is the admitted one, jail says it is this run's machine — and only all three authorise the signal.

- New verdict **`HANDLE_NOT_SHOWN_TO_BE_OURS`**: a handle was taken, and what it names is not shown to be this run's. Handle closed, nothing sent, jail kept, `already_gone` stays `False`. **Nothing branches on `refused_because`**; the distinction is composed once and read by a person.
- **Liveness is asked first.** A guest that ended on its own now reads `PINNED_ENDED` → `HANDLE_GONE` → `already_gone: True`, which is the correct account of it. That is the case the old code got wrong.
- **Unreadable stays unknown.** Every way the read can fail that is not the process ending lands on `HANDLE_UNKNOWN` with the jail kept. `PINNED_UNREADABLE` is a standing of its own, not a missing number.
- **A host with no `/proc` is told apart from a process that is not there.** Both arrive as `ENOENT`. `_the_process_table_is_mounted()` asks `/proc/self`, which is there by construction wherever `/proc` is mounted; without that guard every number on such a host reads as gone and the jail drops behind every live machine.
- `chroot_dir` is **keyword-only with no default** on `_a_handle_pinned_to` and `_stop_the_process_this_run_identified`. A call site that was never updated is a `TypeError`.
- Both stop paths — the deadline at `:2383` and the cleanup after a failed launch at `:2962` — come through `_stop_the_process_this_run_identified`, so there is one contract, not two that drift.
- **`_may_signal_now` is deliberately unchanged.** It holds no reference and is inside no interval; a reading there would look answered and stay racy.
- The recheck is **one-way**: it can turn a signal into a refusal or into "already gone"; it cannot turn a refusal into a signal. So a host where the binding is weaker than argued is no worse off than before.

### The four acceptance questions

| | case |
|---|---|
| owned, confined, live → still stoppable | `d1` — `HANDLE_TAKEN`, one `SIGKILL` through the pinned descriptor |
| replaced / unreadable → cannot signal an unrelated process | `d2` (positive mismatch), `d3` (unreadable) — 0 signals, handle closed, jail kept |
| positive already-stopped cleanup still works | `d4` — `HANDLE_GONE`, `already_gone: True` |
| unknown stays unresolved, not `already_gone` | `d2`/`d3` assert `already_gone is False` explicitly |

Plus `d5` (the jail cannot be left out — signature is `KEYWORD_ONLY`, default is `empty`, both functions raise `TypeError`), `d6` (the reference is opened **before** the handle, asserted by call order), and `d7`/`d8` for the pinned readings themselves.

**Non-vacuity without a mutation campaign.** `d2` is `d1` with one input changed: same child, same start time, same stood-in descriptor, the handle is taken in both, only the jail differs — and the verdict flips from stopping to refusing. `d6` covers what no verdict assertion can: a version that opened the reference *after* the handle would produce the same verdicts and be the arrangement this change exists to avoid.

### Evidence

Every test file that imports the module, my worktree, venv python, at this head:

```
277 passed, 3 skipped in 21.99s
```

`mypy core/agentic_v2_first_boot.py --ignore-missing-imports` reports no error in this file.

**Existing cases were not left alone this time, and that is worth being explicit about.** The rewrite adds a *second* kernel touchpoint to the stop path — the `/proc/<pid>` open — and every stand-in that invented a PID and replaced only `os.pidfd_open` began reading that invented PID as gone, because `/proc/4242` genuinely is not there on this box. Cases in all four pre-existing test files went red that way. The production behaviour was right and the simulations were a transport behind; the repair is in the stand-ins (`_a_stand_in_guest_with_an_identity` in two files, `_a_handle_whose_start_time_reads_back`, and the x13 helper), which now answer for the pinned reference as well as for the handle. No assertion was weakened to accommodate it — the verdicts each case asserts are unchanged.

It was caught by running the suite, not by reasoning, and it was confirmed to be a regression of mine rather than a pre-existing failure by running the same file in a pristine detached checkout of the previous head, where it was green.

### The host delta — not duplicated here

`os.pidfd_open` on this box raises **`ENOSYS`** (measured; kernel `3.10.102`), so the real handle path never runs in any of this. Every case above stands a descriptor in for it, which makes this evidence about the **seam** — the order of the questions, which verdict each answer produces, what is and is not sent — and not about the kernel.

What *was* measured here, with a real child and a real pinned directory descriptor: a live task reads `R` and its `root/.` matches `/`; a zombie still answers `stat` with state `Z` but its `root/.` gives `EACCES`; a reaped task gives `ESRCH` on both. `os.listdir` on a reaped pin returns `[]` with no error, so it is not a usable liveness probe and is not used as one.

Two facts remain out of reach here and are named rather than approximated:

1. that an open `/proc/<pid>` directory on a modern kernel behaves as `fs/proc/base.c` describes, and
2. that `pidfd_send_signal` reaches the process the descriptor names.

Both belong to a host that has the calls. `d1` deliberately does not assert `signal_target`: it is read from `/proc/self/fdinfo`, and a stood-in descriptor has no `Pid:` line.

**No test here signals a process it did not create.** No guest was booted, no VM was started, no model was called.

### Review

**B is the sole independent reviewer; I am not merging this.** The correction above supersedes the argument in the earlier handoff comment on this PR.


hyeonsangjeon:
## Implementation handoff — corrected. The first version of this rested on a false premise.

> **Correction, 2026-09-16.** The version of this comment you may have already read said the whole design rests on "the kernel holds the number while the descriptor is open," and listed that as a fact for your host to confirm. **It is false, and there is nothing to confirm.** Linux 6.17 `kernel/pid.c`: `free_pid` returns the number to the allocator with `idr_remove` *before* it releases the reference with `call_rcu(&pid->rcu, delayed_put_pid)`. A pidfd keeps the `struct pid` alive; it does not reserve the number. Please do not spend a host window on it, and please do not treat Q1 as closed on the old argument.
>
> What it cost was **verdicts, not signals** — `pidfd_send_signal` always delivered through the handle's own `struct pid`, so the target was never at risk. The readings *around* it were the problem: after a reap, `/proc/<number>` could be a successor, so a clean ending could be filed as `not_shown_to_be_ours`, the jail kept, and your run stopped with nothing actually wrong.
>
> The code now binds those readings to an open `/proc/<pid>` **directory** taken *before* the handle (`fs/proc/base.c`: `proc_pid_make_inode` saves `get_task_pid()`, and `get_proc_task()` resolves through it, so the directory is bound to a task the same way a pidfd is). A read through it that still answers proves the reference's task held the number across the moment the handle was taken. **What is below reflects that version.**

Nothing new to build. Your normal exercise already crosses both call sites; this is what to pin and what to read out of it.

### Exact source fingerprint

| | |
|---|---|
| PR / head | #596 / `1b550344061f4a0dbaa793be80531da79bbc1315` |
| base | `main` at `517a1bb2be51970cbab55c4775fbe794ff16d562` (your #594 is in it) |
| `batch-runner/core/agentic_v2_first_boot.py` | blob `4453f806ea71a28a5b17f0ead7323a14013b30db`, sha256 `1bf5f414e26ebec623caa41b8f41a9ee0ba87fa0025ee540fff25c152d319361` |
| `.../tests/test_v2_confinement_is_reconfirmed_while_the_handle_is_held.py` | blob `8f796e71ff00604d78489996d8543d4963223fe9`, sha256 `0541dc633d2b5a6483db2ace82967d096eda363432e37adc3ac5f9cee928cc78` |

Seam: verdict at `:960`, the pin at `:1098`, the reading through it at `:1176`, `_a_handle_pinned_to` at `:1226`, `_stop_the_process_this_run_identified` at `:1457`. Call sites `:2383` (deadline) and `:2962` (teardown). **All of these moved from the numbers in the first version of this comment.**

### The two facts I could not measure

`os.pidfd_open` raises `ENOSYS` here (kernel `3.10.102`), so every local case stands a descriptor in for it. These two are the delta — note that **#1 is not the one the first version named**:

1. **An open `/proc/<pid>` directory keeps addressing its original task on a kernel that has these calls.** This is what replaced the false claim. It is stated by `fs/proc/base.c` rather than guessed, so it is not a hypothesis I am asking you to test — it is named here only so the boundary of what ran locally is honest.
2. **`pidfd_send_signal` reaches the process the descriptor names.** Your record already answers this: `stop_signal_target` (set at `:2411` on the deadline path, `:2968` on teardown) is read from `/proc/self/fdinfo`, which a stood-in descriptor has no `Pid:` line for. On a real handle it should equal the pid that was pinned.

### What to read out of an ordinary run

**Deadline, guest still running.** The happy path is unchanged: `outcome: stopped_by_the_deadline`, `stop_handle: by_a_pinned_handle`, jail destroyed. The thing to check is that the confinement reconfirmation did **not** fire on a machine that was genuinely yours — if it did, you get `outcome: overran_and_was_left_alone` with `stop_handle_verdict: not_shown_to_be_ours` and `left_alone_because` naming your own jail. That is the expensive failure direction and the one worth looking for first.

**Deadline, guest ended on its own.** `outcome: booted`, and internally `HANDLE_GONE` with `already_gone: true`. **This is the case the old code got wrong** and the one the correction above is really about: before the pin, a reaped guest whose number had been handed on read as a live unconfined stranger, so an ordinary clean ending came out as `not_shown_to_be_ours` and stopped your run. If you still see `not_shown_to_be_ours` here, the liveness-before-identity order broke — which is what `d4` pins.

**Teardown after a failed launch, process already stopped.** `confirmed_stopped: true`, `left_alone_because` saying it stopped on its own. Unchanged.

`refused_because` distinguishes a positive mismatch ("not the jail this run made and holds") from an unreadable one ("nothing can be shown to be confined to it"). **Do not branch on it** — it is for a person. Branch on `stop_handle_verdict` / `signalled` / `confirmed_stopped`.

### One new thing worth knowing about your hosts

The stop path now opens `/proc/<pid>`, which it did not before. A host where `/proc` is not mounted would read **every** number as gone and drop the jail behind a live machine, so `_the_process_table_is_mounted()` asks `/proc/self` on the error path and turns that back into `unknown`. I do not expect this to matter on your hosts — it is here so that if you ever run the backend somewhere unusual you know the guard exists and why.

### Not asked for

No re-run of the six pidfd verdicts, no re-run of your guest/deadline exercise beyond what you were going to do anyway, no second VM window, and — after the correction above — **no host time spent on the number-reservation question at all.** If your next normal exercise already produces the records above, that is the delta; share it here rather than arranging anything new.

hyeonsangjeon:
## Head moved to `422f750edb6466ca2dbdcf8ee7193120a5c7f488` — `pytest` repaired, plus one test that did not exist before

One commit since the head you last saw. **Test-side only: one file, +63/−1. No production line moved**, so the seam map in [`5691093582`](https://github.com/hyeonsangjeon/gdpval-realworks/pull/596#issuecomment-5691093582) still holds line for line.

All three checks are green on this sha. `pytest` finished `12289 passed, 61 skipped, 46 deselected in 1332.01s`.

**The title changed too, and it is not cosmetic.** It read *"re-ask confinement inside the interval where the PID cannot move"* — the falsified premise itself, and the string that would have become the squash-merge message on `main`. It is now *"re-ask confinement of the process the handle names, not of its number"*. No commit, so CI is unaffected.

### What was red, and why my local run could not have caught it

`pytest` at `1b550344` was `1 failed, 12270 passed, 61 skipped, 46 deselected in 1291.58s`. The one failure, verbatim:

```
tests/test_v2_ownership_is_bound_to_the_launch.py:929: in test_x12_...
    assert "already gone when a handle was asked for" in why_not
E   AssertionError: assert 'already gone when a handle was asked for' in
    'process 14213 was already gone when this run reached for a reference to it,
     so there is nothing to stop'
```

`x12` is gated on a kernel that actually has `pidfd_open`. This box is 3.10.102, so it skips here and had never once executed against my change — the local green was silent about it, not confirming. On the runner the gate opens: the green run's `-rs` report carries no pidfd skip reason at all, so `x12` and the whole `_NEEDS_A_HANDLE` family did run there.

The behaviour was right and the expectation was stale. Under the reordering, a pid that has already been reaped fails at `os.open("/proc/<pid>", O_DIRECTORY)` and **never reaches `pidfd_open` at all**, so the refusal it produces is the pin's (`core/agentic_v2_first_boot.py:1155`) and not the handle's (`:1340`). I did not restore the old string: on that path no handle is asked for, so the old wording would have named a moment that did not happen.

The fix asserts the pin-site wording **and** asserts the handle-site wording is absent, so a later reordering that silently sends reaped pids back through `pidfd_open` fails here instead of passing quietly.

### The gap that opened underneath it — `x14`

With `x12` moved to the pin, `grep` found `:1340`'s message asserted by **no test at all**. My reorder had left that branch reachable only by a genuine race: alive when the reference was taken, reaped a moment later.

`x14` arranges that race. The child is real and alive, the pin is the real one, and only `os.pidfd_open` is stood in for — which is also the only way a box without pidfd reaches that line, so unlike `x12` **this one runs everywhere**. It asserts the message, the `HANDLE_GONE` verdict, and that the pin fd is closed on the way out of that refusal, using the same fd-count idiom already at `:921`/`:976`.

Both moments are `HANDLE_GONE` and both mean the machine is gone. What differs is when the run found out, and a record naming one while meaning the other points at the wrong moment.

### One thing that red run bought us

The corrected handoff listed the reaped case as not measurable from here. **It was measured — by the failure itself.** On the runner's real kernel a reaped pid returned `HANDLE_GONE` through the new pinned-reference route, with `pidfd_open` never called. Its first two legs (`nothing is None`, `verdict == HANDLE_GONE`) passed there; only the substring did not. That is real-kernel evidence for the new ordering on the death path, and it is not something I could have produced locally.

It is still not a guest boot, and it says nothing about whether a real Firecracker process behaves this way under your host. That remains yours.

### Not done

No production file touched in this commit, no self-merge, no follow-up PR. Review the delta only — `1b550344..422f750e` is the whole of it.

hyeonsangjeon:
Independent review of the corrected implementation at `422f750e` — not of the
commit message's account of it, and not of the withdrawn number-reservation
premise. Three areas, as scoped.

**1. Pin-before-handle binding — holds, and the order is asserted rather than inferred.**

`_a_proc_reference_pinned_to` is called at `core/agentic_v2_first_boot.py:1330`,
`os.pidfd_open` at `:1334`, and every reading afterwards goes through the
descriptor: `os.open("stat", dir_fd=pinned)` at `:1197`, `os.readlink("root",
dir_fd=pinned)` at `:612`, `os.stat("root/.", dir_fd=pinned)` at `:668`. The
kernel argument the order rests on is the right one — `proc_pid_make_inode`
takes `get_task_pid()` at lookup and `get_proc_task()` resolves every later read
through that saved `struct pid`, while the number is returned to the allocator
by `free_pid`'s `idr_remove` at reap. So a read that answers at t3 says the
reference's task was unreaped through t3; a task holds its number until it is
reaped; the span therefore contains t2; one number has one holder. The handle
and the reference are the same process.

`tests/test_v2_confinement_is_reconfirmed_while_the_handle_is_held.py:473`
asserts `order == ["reference", "handle", "confinement"]` positionally, so
reversing the two opens fails the suite rather than producing the same verdict
by a racy route. That is the part that makes this reviewable.

**2. Owned signal and unknown cleanup — every non-positive answer refuses and keeps the jail.**

`PINNED_UNREADABLE` lands `HANDLE_UNKNOWN` with the jail held (`:1378-1387`).
`already_gone` is reached only from positive findings — `ProcessLookupError`
from `pidfd_open` (`:1337`) and a start time that reads back different
(`:1393`) — never from a reading this run could not take. The four remaining
verdicts refuse without signalling (`:1548-1560`), `HANDLE_UNSUPPORTED` now
among them. No delivery path by number survives: the only `os.kill` left in the
module is the signal-0 liveness probe at `:883`. Descriptor hygiene is complete
— `finally: os.close(pinned)` at `:1430`, `finally: os.close(handle)` at
`:1526`, and a close on each early return between them.

Two things I checked because they would have inverted the safe direction and
did not: `A_STAT_LINE_FITS_IN` is 8192 against a ~300-byte line with `comm`
capped at `TASK_COMM_LEN`, so no truncation can forge a start-time mismatch into
a false `HANDLE_HANDED_ON`; and `_the_reference_can_be_read_here` is a real
`os.supports_dir_fd` probe whose `False` refuses rather than falling back to the
path form.

**3. Both callers — one contract, and the deadline case is the one that was broken.**

Both reach `_stop_the_process_this_run_identified` with `chroot_dir` required
(`TypeError` if omitted, asserted at `tests/...:395-415`). On the deadline path
a reaped-and-recycled number now reads its start time back through the
reference, lands `HANDLE_HANDED_ON`, and ends `already_gone=True` → `booted` —
which is the behaviour the old arrangement got wrong.

One observation on the cleanup path, **not a blocker and not in this delta**:
its pre-stop readings at `:2922` (`_still_running`) and `:2944`
(`_confined_to_this_runs_jail`, no `pinned=`) are still by number, so the same
reaped-and-recycled case refuses there before ever reaching the repaired code —
`confirmed_stopped` stays `None`, the jail is kept, nothing is signalled. That
is the conservative direction and matches what the docstring says the path form
is for, so it stays as it is. Recording it so a later reader does not mistake
the repair for covering both entry points.

One non-blocking test note: `d7`'s stand-in at `tests/...:452` is
`reading(pid, chroot_dir, **named)` and records the call without asserting
`named` carries `pinned`, though the docstring claims the reading "has to go
through it". A one-line `assert named.get("pinned") is not None` would make that
second claim load-bearing too.

**Evidence reused rather than re-run.** Required CI is green at this exact head
(35059723605 / 35059723604 / 35059723596): 12289 passed, 61 skipped, 46
deselected. The kernel-gated cases matter here — `test_v2_only_stops_a_process_
it_started.py` and `test_v2_ownership_is_bound_to_the_launch.py` show three
skips on this 3.10.102 box and **zero** on the runner, so the pidfd paths
actually executed on a 6.x kernel. Locally: 133 passed, 3 skipped across the
five changed files. No host battery re-run and no mutation campaign.

Repair count across rounds for this defect: 2 — `7e7d48b` (withdrawn premise),
`b38b218` (this mechanism). Accepted at the second.

PR #597, opened 2026-09-16
feat(azure): the rbac verdict says which workloads the missing role gates

hyeonsangjeon:
## What this changes

`scripts/azure_rbac_diagnostic.py` ends on a verdict —
`role_missing_and_an_owner_must_grant_it` — and that reads as a blanket block on
every Azure AI call from this identity. It is not one, and the report never said
so. This adds the missing sentence, measured rather than asserted.

```
verdict              : role_missing_and_an_owner_must_grant_it

The role is missing and this identity cannot assign roles. An Azure Owner on
the account or project has to run the command below.

what this role gates, and what it does not:
  stopped by it        : code-interpreter -> project
  addressed elsewhere  : grader -> direct-v1, inference -> direct-v1, narrative -> direct-v1

A workload addressed elsewhere is not waiting on this role. A workload stopped
by it is refused with http 403 after the token check ahead of it passes, which
is what it did to exp032.
```

## Why it is right

`AzureAIRouteSettings.select()` (`core/azure_ai_clients.py:466-496`) sends
`CODE_INTERPRETER` to the project endpoint. Every other workload under
`direct-v1` or `project-ci` resolves to the account-scoped `direct-v1` endpoint,
which the `Cognitive Services OpenAI User` role this identity **already holds**
serves. `scripts/run_agentic_v2_stage.py:1074` leases
`AzureAIWorkload.INFERENCE` — so the paid V2 stage is not waiting on the owner.
`exp032` is the other half of the same fact: its token check passed and then all
five tasks took `PermissionDeniedError (http 403)` from the project route.

I designed a `--require-project-role` gate on the paid job before measuring
this, and discarded it — it would have refused runs that work.

## How it cannot drift

- The section **calls `select()`**; it does not restate the routing rule.
- `test_every_typed_workload_is_accounted_for_exactly_once` iterates
  `AzureAIWorkload`, not a hand-written list, so adding a workload without
  classifying it is a red test in the commit that adds it.
- A route that cannot be built reports `not_measured_because` instead of
  reporting that nothing is gated; `workloads` is seeded in the report dict so a
  run that stops early still carries the key.
- `test_a_route_error_cannot_publish_an_identifier_through_the_new_section`
  forces `from_env` to raise with an endpoint in the message and asserts the
  existing `redact` / `leaked_placeholders` pass still fails closed.

Two overlays are applied to a **copy** of the environment before the route is
built: the profile defaults to `project-ci`, and
`AZURE_AI_REQUIRE_EXPECTED_IDENTITIES` is forced to `0`, because this
diagnostic's workflow deliberately omits the expected account and project names
to keep them out of a public step header. The real environment is passed through,
so `_reject_static_azure_credential_env` still applies.

## The external block, stated once

Not fixed by this PR and not fixable from here:

| | |
|---|---|
| principal | VM managed identity `8544cbfc…` (`gdpval-devhost-vm`) |
| its tenant | `6d93cc9b…`, subscription `4b7c60a5…` |
| target | account `hjeon-fdpo-foundry-eus2`, project `gdpval-realworks`, deployment `gpt-5.4` |
| roles missing | `Foundry Agent Consumer` `eed3b665-ab3a-47b6-8f48-c9382fb1dad6`, `Foundry User` `53ca6127-db72-4b80-b1b0-d745d6d5456d` |
| measured error | `principal_can_assign_roles: false`, `read_failures: []`, run `35053114735` |
| why no assignment was attempted | the target account returns 0 Resource Graph matches in the only reachable tenant, and its endpoint answers `HTTP/2 401` — it is **cross-tenant**, so a role assignment from here is not a permission problem, it is not expressible |

Cross-tenant RBAC was not invented or attempted. A human tenant owner is
genuinely required for the Code Interpreter route; nothing else is.

## Verification

- `tests/test_azure_rbac_diagnostic.py` — 88 passed (80 before; 8 added).
- `test_azure_ai_route_preflight.py` + `test_azure_ai_clients.py` +
  `test_run_agentic_v2_stage.py` — 305 passed.
- `mypy scripts/azure_rbac_diagnostic.py` — one pre-existing error in
  `core/cost_metering.py:138`, an untouched file.
- The workflow's forbidden-identifier grep is clean against the modified source.
- No workflow change needed: `azure-rbac-diagnostic.yml` already runs the
  script, so the section is emitted on the next dispatch.
- Model calls 0, paid runs 0. No VM was started for this change.


hyeonsangjeon:
## Independent review — sound, merging at `a11427ed25037c6f76117a5ff7b45cf27c19961f`

Two files, +301/−0. `pytest` and `advance-check` SUCCESS on this sha, `mergeStateStatus: CLEAN`. No `freeze-check` — nothing here touches the grader corpus, so its absence is expected and not a missing check.

I did not re-run your suites and did not dispatch the diagnostic workflow. I checked the two things a route-reporting change can get wrong: whether it can move paid authentication, and whether the new sentence can print something the log must not carry.

### It cannot alter what the paid path enforces — three independent reasons

- `AzureAIRouteSettings.from_env(overlaid)` passes an explicit mapping, so the `values = os.environ if env is None else env` guard at `azure_ai_clients.py:344` means `os.environ` is never read and never written on this path.
- `overlaid` is a fresh `{**env, ...}`, so the caller's mapping is untouched. `test_the_identity_assertion_is_switched_off_without_touching_the_caller` pins that.
- That test only covers the caller's dict, so I checked global state separately: `core/azure_ai_clients.py` contains no `os.environ[...]` write, no `setdefault`, no `putenv`, no `global`, no `lru_cache` and no module- or class-level memo. `from_env` is a classmethod returning a fresh dataclass and `select` reads only `self`. There is nothing for a later paid call to pick up.

Reachability agrees: `which_workloads_this_role_gates` has exactly one call site, `azure_rbac_diagnostic.py:741`, inside `diagnose()`. The only non-test consumer of the module is `scripts/azure_boot_host_survey.py`, which imports it for its helpers and builds its own report shape (`location_asked_about`, `host_definition`, `findings`) — so the new `workloads` key breaks no existing reader.

### The privacy question — I checked the raise sites, not the docstring

`unroutable` writes `str(broke)` straight into `"    - {workload}: {because}"`, and your one privacy test monkeypatches **`from_env`**, which lands in the `not_measured_because` branch and never exercises the `select()` branch. So I went at it from the other end and enumerated what those two calls can actually raise.

- `select()` — four reachable raises, all fixed literals (`"Code Interpreter requires the project-ci profile"`, `"Foundry project endpoint is unavailable"`, `"direct v1 endpoint is unavailable"`, `"legacy endpoint is unavailable"`). The `AzureAIWorkload(workload)` coercion cannot fire here because the argument is already an enum member from `for workload in AzureAIWorkload`.
- `from_env()` — every raise names an environment **variable**, never a value: `_env_value`, `_reject_static_azure_credential_env` (joins names), `_classify_typed_endpoint` (`f"{variable} must contain a {required_kind.value} endpoint"`), `_deprecated_endpoint_error` (names plus an enum value), `_validate_account` / `_validate_project` / `_validate_hostname`, and the two profile checks.
- `classify_endpoint` is the one that receives raw URLs and it raises sixteen times, every one a constant string.

So `str(broke)` and `{broke!r}` can only ever carry a fixed sentence. The branch your test misses is safe for a stronger reason than redaction, and the redaction net behind it is real and covers both output forms: `main` serializes prose **or** `--json` into one `body`, runs `redact` over the whole string, and returns 3 without printing if `leaked_placeholders` finds a survivor.

### The `project-ci` default does not change the claim that matters

`ROUTE_PROFILE_ENV: env.get(...).strip() or "project-ci"` only fills an empty variable, so a set profile wins. And the load-bearing result is profile-robust: under `project-ci` and `direct-v1` every non-code-interpreter workload takes `direct_v1`, under `legacy-rollback` it takes `legacy`. In no profile does `inference` land on `PROJECT`, which is what `run_agentic_v2_stage.py:1074` leases. What does move is `code-interpreter`: under a non-`project-ci` profile `select()` raises and it appears under *no route at all* rather than *stopped by it*. That is an honest answer rather than a wrong one, and the new renderer prints the reason.

`test_every_typed_workload_is_accounted_for_exactly_once` is the test that keeps this honest as the enum grows — a new workload cannot quietly vanish from the split.

### One note, not a blocker, and not caused by this PR

`LEGACY_ENDPOINT_ENV` (`AZURE_OPENAI_LEGACY_ENDPOINT`) is read by the route layer at `azure_ai_clients.py:68` but is absent from `SENSITIVE_ENV` in `core/azure_control_plane.py:62–73`, which carries the other three endpoint variables. Nothing leaks today, because nothing on this chain interpolates a raw endpoint. But it means the redaction net has a hole in exactly the place a future message in this chain would land, and `leaked_placeholders` would not fail closed on it. Your file, your call whether it is worth a line.

### Not done

I did not re-run `test_azure_rbac_diagnostic.py`, did not run the diagnostic workflow, did not touch either file, and opened no follow-up PR. Merging now; #596 is still yours to review and is unaffected by this.

PR #598, opened 2026-09-16
feat(v2): make a same-host isolated stage run actually reachable

hyeonsangjeon:
The paid V2 stage has accepted `--isolated-approval` for some time and nothing
in the repository produced one. Every dispatch therefore resolved to
`AgenticV2FixtureBackend` no matter what `isolation` said, and the only
approval that ever existed was written by hand on `gdpval-devhost-vm` — a
machine that has the hardware virtualisation but not the federated session.

This makes the isolated path reachable from the job that already holds the
session, by putting the guest on the same runner rather than by moving proof,
credentials or roles between hosts.

## What changes

**One bounded capability check, first.** `scripts/check_runner_can_host_a_guest.py`
reads `/dev/kvm`, the nested-virtualisation flag and the launcher programs on
the runner and reports. Two dispatches have already answered it — runs
`35072131325` and `35072747069` — and `capability-check` is a dispatch option so
the question can be re-asked without a cohort attached.

**A composite action both jobs call.** `.github/actions/boot-a-guest-on-this-runner`
installs the pinned launcher by digest, re-reads the capability now that the
programs are there, boots one guest under the jailer, and writes the approval —
but only if the real `select_backend` accepts it first. The free job and the
paid job call the same action, so "the dry run proved it" means the arrangement
the paid job will actually have, not a copy of it that can drift.

**The flag, passed or the run stops.** A `same-host` dispatch adds
`--isolated-approval` and fails the step if the boot left no approval. There is
no fallback to the fixture: a run that reports isolation it never had is the
outcome this whole path exists to prevent.

**Root, and only on that dispatch.** `core/agentic_v2_first_boot.py` execs the
jailer in-process, and the jailer chroots, makes the device node and writes
cgroup files. So the stage process itself is root on a `same-host` dispatch. A
`fixture` dispatch is unchanged down to the command line.

`sudo env NAME=value` rather than `sudo -E`: `-E` needs a SETENV tag that
cannot be confirmed without dispatching, and it carries the whole environment,
where naming each variable makes an omission a failure rather than a surprise.

**The session root would otherwise not see.** `sudo` gives root a different
`$HOME`, and `DefaultAzureCredential` lands on `AzureCliCredential`, which reads
`$AZURE_CONFIG_DIR` or `$HOME/.azure`. `AZURE_CONFIG_DIR` is pointed at the
directory `azure/login` wrote, and a free preflight asks the same question
through the same mechanism before the dataset is fetched — the one thing about
this arrangement the free rehearsal cannot otherwise show. Nothing is copied
and nothing is exported.

## What it deliberately does not do

No new self-hosted runner, no new federation or role grant, no credential
export, no public tunnel, and no implicit non-isolated fallback. The proof is
made on the runner that consumes it; `scripts/write_isolation_approval.py`
refuses an artefact whose boot id is not this boot of this machine, because two
runners built from one image report the same kernel release and kernel equality
would not notice one carried between hosts.

The boot runs *before* `azure/login`, so nothing in the boot path is on the
machine at the same time as a federated session. The guest gets a work disk and
nothing else.

## Evidence

* `tests/test_v2_same_host_isolation_is_fail_closed.py` — 17 → 37 tests.
* `tests/test_run_agentic_v2_stage.py` — 62 tests; the stage detector changed
  with the workflow (see below).
* Ten planted defects in the same-host wiring, ten caught. Five planted
  defects in the dataset/parquet wiring, five caught. Every plant gated on the
  file still parsing as YAML, and the file restored byte-for-byte afterwards.
* V2 slice: `2871 passed, 24 skipped, 9597 deselected`.
* `mypy` clean on the six scripts, including the two this workflow gates on.

**One test change worth a reviewer's attention.** `INVOKES_THE_STAGE` required
the literal word `python` in front of the script path. That stopped being true
once the paid job runs `"${RUN[@]}" scripts/run_agentic_v2_stage.py`, and the
detector then counted one job where it asserts two — it did not fail, it
quietly stopped checking the paid job. Broadening it far enough to see that
made it match `python -m mypy scripts/run_agentic_v2_stage.py`, which
type-checks the stage rather than running it and has no business fetching a
dataset. It now excludes lines that run a module. Both directions are covered
by the five plants above.

## What this does not yet prove

A `same-host` dispatch has not been run. The capability checks say the runner
opens `/dev/kvm`; they do not say a guest boots on it, that the approval is
admitted, or that `exec_run` makes a round trip. That is the next dispatch, on
`advance_check_5` with its existing manifest, and it is what this PR exists to
make possible.


hyeonsangjeon:
## One commit added after this was opened, and a note on the reviewer field

**`d038897` — ask for a token as root only after nothing else needs the session.**

Found by reading the paid job end to end while CI ran, not by a failing test.
The `sudo` credential preflight sat above two steps that use the same Azure
session as the job's own account: `Verify Azure OIDC session identity` and
`Validate the Azure AI route`. Acquiring a token is a write —
`verify_session_identity` runs `az account get-access-token`, and `az`
refreshes and rewrites its MSAL cache. A file rewritten by root inside a
`runner`-owned directory comes back root-owned, so the route validation could
have failed on a permission error in a step that has nothing to do with
privilege.

The preflight now runs below both of them. Ordering is the whole mitigation:
there is no copy of the credential store, no second config directory and no
`chown` of one. `test_the_same_question_is_asked_at_the_privilege_that_will_ask_it`
now pins the relation, and I proved the new assertion load-bearing by actually
swapping the two YAML blocks (caught; file restored byte-for-byte).

Wider slice after the reorder: `3713 passed, 24 skipped, 8755 deselected`.

**The reviewer field cannot be set.** A and B share one account, so the API
refuses:

```
POST .../pulls/598/requested_reviewers
422  Review cannot be requested from pull request author.
```

`gh pr edit --add-reviewer` exits 0 and prints the PR URL without adding
anything, which is worth knowing — it would let someone claim a review was
requested when it was not. So this comment is the handoff: **A, the head to
review is `d038897`.** No review has happened yet and none is recorded.

hyeonsangjeon (commented):
Reviewed at `d038897` against `decfed2`. I had built the same integration on a
parallel branch and did not open it — this one is the better of the two on the
points where they differ, so I converged here rather than competing. Two
comments below, neither blocking.

## What I checked, and what held

**Process identity, termination, cleanup.** Unchanged by this PR. The only edit
to `run_agentic_c2_first_boot.py` is +19 lines recording `boot_id`; the launcher,
the jailer handoff and the teardown paths are untouched, so the 22/22 wiring
check at `57fa6af` still describes them. What *is* new is the cleanup surface the
runner adds, and it closes: the approval is written by root and `chmod 0644`'d
before the unprivileged upload step reads it; `agentic-v2-run` is `chown -R`'d
back after a root-run stage, before collection; the jail scratch is root-owned
under `/var/tmp` on a runner that is destroyed at the end of the job.

**The host binding.** Keying on `/proc/sys/kernel/random/boot_id` rather than on
kernel release is the right call and strictly stronger — my branch used the
release alone, and two runners out of one image report the same string, which my
own docstring had to concede. Keeping the release check alongside it costs
nothing and is kept. Absent is refused rather than skipped, which is the 2026-09-14
lesson applied in the right direction.

**Writing the approval through the real `select_backend`.** An approval that
would be refused on task one is refused here instead, before `azure/login` and
before the ledger opens. This is the part I'd have got wrong: my version
round-tripped through `IsolationApproval.load` only, which validates the shape
and not the admission.

**Order of boot and login.** Booting above `azure/login` means no federated
session exists on the machine while root is building a chroot and exec'ing a
downloaded binary. That is worth the extra minute.

**`sudo env NAME=value` rather than `sudo -E`.** Correct, and it fixes a latent
defect in my version: `-E` needs a `SETENV` sudoers tag that the runner image
does not grant, so mine would have dropped `AZURE_CONFIG_DIR` and failed to
resolve a credential — several minutes in, as an auth error rather than as a
sudo one. Naming each variable also makes the boundary reviewable, which is the
answer to "what does the control process hand across". Nothing on that list is a
secret; `AZURE_CONFIG_DIR` is a path, and a Firecracker guest cannot read the
host process's environment regardless.

**Fail-closed at both gates.** Free job line 515 and paid job line 876: a
`same-host` dispatch with no approval file stops, rather than appending no flag
and running the fixture under a dispatch that said isolated. I mutated both
guards to `if false` and the parametrised test caught both.

**The tests run here.** 147 passed, **0 skipped**, on this kernel-3.10 box:

```
tests/test_v2_an_approval_is_proved_before_it_is_written.py      18
tests/test_v2_hardware_virtualisation_is_used_not_assumed.py     12
tests/test_v2_the_launcher_programs_are_pinned_not_picked.py     18
tests/test_v2_same_host_isolation_is_fail_closed.py              37
tests/test_run_agentic_v2_stage.py                               62
```

Zero skips is the result worth stating. The 22 pidfd tests skip on this kernel
and a green local run says nothing about them; the injectable `opener`/`ask`/
`closer`/`device` seams in `agentic_v2_kvm_probe.py` are what buys the
difference. Planting `if False:` in place of the `boot_id` comparison turned 3
tests red, so the host binding has teeth and not just coverage.

## Two comments

**1 — the firecracker digest is self-certifying, and it need not be.**
`fetch_release` downloads the tarball, downloads `…tgz.sha256.txt` from the same
release in the same moment, and compares one against the other. That catches a
truncated transfer. It does not catch a re-uploaded asset, because whoever could
replace the asset could replace the digest beside it — which the module docstring
says accurately, in those words, rather than writing "verified" and leaving it.

This is not below the repo's existing bar: `fetch_pinned_kernel` checks
`size_bytes` and nothing else on a first download, because upstream publishes no
digest for the kernel at all. But firecracker differs in the one way that
matters — upstream *does* publish a digest, so a committed constant costs one
line and converts "unchanged in transit" into "the file this branch was reviewed
against". It also matters more than the kernel does: this binary is exec'd as
root on the runner, in the job that holds the federated session.

I fetched the published digest today and it is

```
59450b9171ff2ebdf2f9a25be3a248a7ba79fb6371aec51a9d6d8eefca7b4faf  firecracker-v1.13.1-x86_64.tgz
```

so adding `"sha256": "59450b91…"` to `PINNED_FIRECRACKER_RELEASE` and refusing
when `wanted != release["sha256"]` before the existing comparison is a two-line
change against a verified value. Non-blocking: the current check is honest about
its own reach, and a moved v1.13.1 asset would be a different and larger
incident. Worth doing in a follow-up.

**2 — the capability probe asks an account that never boots a guest.**
`allow_this_job_to_open_kvm.sh` grants `/dev/kvm` to the job's account, and the
reading that gates the dispatch (free job line 297, and the action's "Read the
runner again") runs as that account. Every process that actually touches the
device — `first_boot`, the jailer, firecracker, and the stage itself — runs as
root under `sudo`, where the ACL is irrelevant.

That is stricter than necessary rather than wrong: root can open the device
whichever way the grant went, so no false admission is possible. The path that
does exist is a false refusal — the script `exit 0`s on total failure by design
("the capability reading below is the answer"), so if both `setfacl` and `chown`
miss, the probe answers `not_permitted` and a `same-host` dispatch stops on a
runner root could have used. It costs a free job, not money, and the reading
prints `device_owner` so the log points straight at the cause. Asking as root
and dropping the grant would collapse it, but the current arrangement exercises
the grant on every dispatch, which has its own value. Your call.

## What none of this establishes

Nothing here has run on a real runner. Every claim above is about the code and
the local suite; whether `ubuntu-latest` exposes working nested virtualisation is
the fact the whole `capability-check` mode exists to answer, and it is unanswered
until a dispatch answers it. `KVM_CREATE_VM` over `KVM_GET_API_VERSION` is what
makes that answer trustworthy when it arrives — the version ioctl succeeds on a
host where nesting is off, which is exactly the case that would otherwise be
found out inside a jailer, minutes later, in a cohort that had already been paid
for.

Merging on green `pytest`; `advance-check` and `freeze-check` are already past.

PR #600, opened 2026-09-16
fix(v2): take the program out of the release, not its debug symbols

hyeonsangjeon:
## What happened

A same-host dry run on a GitHub-hosted runner — [run 35082237811](https://github.com/hyeonsangjeon/gdpval-realworks/actions/runs/35082237811), and again on [35082315150](https://github.com/hyeonsangjeon/gdpval-realworks/actions/runs/35082315150) — failed at the first step of `boot-a-guest-on-this-runner`, before anything was booted:

```
refused: looking for one firecracker in firecracker-v1.13.1-x86_64.tgz found
['release-v1.13.1-x86_64/firecracker-v1.13.1-x86_64',
 'release-v1.13.1-x86_64/firecracker-v1.13.1-x86_64.debug'].
The archive is not laid out the way this release was
```

The archive is laid out exactly the way this release is. v1.13.1 ships each
program twice: the binary at mode `0755`, and its detached debug symbols under
the same name with `.debug` on the end at `0644`. Both answer to the prefix
`firecracker-`, `unpack_programs` requires a prefix to match exactly one, so the
installer refused the correct release.

This is why nothing caught it: `the_usual_archive` — the fixture whose docstring
said "shaped the way upstream ships one" — contained no `.debug` members, and
`an_archive` gave every member mode `0755`. The fixture could not express the
shape that breaks the code.

## The change

`unpack_programs` now narrows by prefix and then chooses among the candidates by
the executable bit. The suffix is not named anywhere: `.debug` is upstream's to
rename, and the thing that actually distinguishes a program from its symbols is
that only one of them is a program.

Refusing none-or-more-than-one is unchanged, and so is refusing two *runnable*
matches — `test_an_archive_offering_two_of_a_program_is_refused` still passes
untouched. If a later release stops marking the binary runnable, the prefix
still matches, the mode filter finds nothing, and the refusal now names the
members it set aside rather than printing an empty list at someone who can see
the file sitting right there.

## Verification

Against the real v1.13.1 tarball, downloaded from the pinned URL:

| | picked | sha256 | upstream `SHA256SUMS` says |
|---|---|---|---|
| firecracker | `firecracker-v1.13.1-x86_64`, 3,284,984 B | `7d3924e0…06f6e` | `7d3924e0…06f6e` ✓ |
| jailer | `jailer-v1.13.1-x86_64`, 2,289,120 B | `7ff76f93…98498` | `7ff76f93…98498` ✓ |

The digests match the entries the archive carries for the *binaries*; the
`.debug` entry is a different digest (`836379a3…42519`). Both extracted files
start `\x7fELF`.

- `tests/test_v2_the_launcher_programs_are_pinned_not_picked.py` — 21 passed
  (17 before; 4 new)
- V2 launcher/boot/isolation sweep — 261 passed, 3 skipped
- full backend suite — see the CI run at this head
- `mypy core/agentic_v2_firecracker_release.py scripts/install_pinned_firecracker.py` — clean

## What this does and does not establish

It does not establish that a guest boots on a GitHub runner. The composite
action failed at step one, so its remaining three steps — read the runner again
with the programs present, boot one guest, write the approval — have never run
anywhere but the dev host.

What the same run did establish, before it failed, is that the host is not the
obstacle. Its capability reading (artifact `runner-capability-same-host`) reports
`could_host_a_guest: true` and `blocked_by: []` on `ubuntu24`, kernel
`6.17.0-1022-azure`: the processor reports `svm`, `/dev/kvm` opened and answered
KVM API version 12, and an empty machine was created and closed again. The one
unmet claim was "the Firecracker programs are installed" — which is the step
this PR repairs.

So the next same-host dry run is the one that answers the boot question, and it
answers it for free.


hyeonsangjeon:
No human review happened on this PR and none is recorded. `gh pr edit --add-reviewer` cannot request one here — the account that opened it is the account that would review it, and GitHub answers `422 Review cannot be requested from pull request author` while `gh` still exits 0. Saying so rather than leaving `reviewRequests: 0` to be read as "nobody objected".

Merging on: three required checks green at the exact head `8ce07c05b3c6b27fa89d66a082a5a7fc32d95a3e` (advance-check 2m17s, freeze-check 17s, pytest 20m33s), the defect is an observed CI failure reproduced on two independent runs, and the fix is checked against the real upstream archive's own `SHA256SUMS` rather than against my reasoning about it. The change is confined to V2 launcher-install code that this session owns.

PR #601, opened 2026-09-16
fix(v2): ask the gate as the account that boots the guest

hyeonsangjeon:
Two `same-host` dispatches stopped on the same step, on two different runners.
Nothing was spent by either.

[35087786168](https://github.com/Hyeonsangjeon/gdpval-realworks/actions/runs/35087786168)
and
[35088055576](https://github.com/Hyeonsangjeon/gdpval-realworks/actions/runs/35088055576):

```
NO   /dev/kvm is present and this process may not open it (Permission denied)
```

## The runner was never the problem

From the first run's own log, same job, same machine, same account:

```
11:00:16  granted: an ACL entry giving runner read and write on /dev/kvm
11:00:16  yes  /dev/kvm opened, answered API version 12, and created an empty
                machine which was closed again
11:01:04  NO   /dev/kvm is present and this process may not open it
```

The grant took — `getfacl` printed `user:runner:rw-` — and the reading that
followed it did not stop at the filename: it opened the device and asked it to
create an empty machine, which it did. Hardware virtualisation works on these
runners. It made a machine at 11:00:16 and was refused at 11:01:04, and nothing
in the forty-eight seconds between so much as names the device.

So the thing that did not survive is the ACL granted to the job's *own account*.

## Why that stops nothing except the gate

Everything on a `same-host` dispatch that opens the device for real is root:

| what | where |
|---|---|
| the first boot, which execs the jailer | action step 3, `sudo env` |
| the approval the selector must accept | action step 4, `sudo env` |
| the stage itself | workflow `780` and `904`, `sudo env` |

Root bypasses DAC, so a 0660 `root:kvm` node opens for all three whatever the
ACL says. The reading between them was the only thing in the path that depended
on the ACL, and it was asking after an account that never opens the device.

## The change

The gate is taken at the privilege that boots. That is not a new rule — it is
the rule already written under the approval step three lines further down in the
same file:

> a gate is worth having only if the process that checks is the process that
> will act … a false refusal that stops a good run is not a safer failure, it is
> a different wrong answer.

**Nothing is relaxed.** The reading still passes `--require-programs`, still
opens the device, still asks it to create an empty machine, and still exits
non-zero when a gated reading fails. It is asked as the user who will do that
for real. The report it writes is `chmod 0644` afterwards because the rehearsal
uploads it as an artifact and it would otherwise be root-owned; it is a reading
of a machine and holds no secret.

The device's mode and ACL are printed immediately before, so the next failure
here says which of the two it was instead of leaving it to be re-derived from
timestamps the way this one had to be.

## A second repair for the same defect exists, and I read it before pushing

A parallel worktree carries a different fix: keep the reading unprivileged and
re-take the grant immediately before it, shrinking the window from forty-eight
seconds to the gap between two adjacent steps. It also adds a `getfacl` to
`allow_this_job_to_open_kvm.sh` *before* the script changes anything, so a
second grant can say what the first one's entry had become. Its tests pin the
new step's position between the install and the reading, and pin that it goes
through the one script rather than an inline `setfacl`. It is careful work and
it would also unblock the run.

Two reasons I think this one is the better repair, and one thing from that one
that is worth having either way:

1. **It does not depend on the mechanism being what we think.** That version
   names its explanation — a privileged step causing logind to re-apply the
   device ACLs it owns — and says plainly that it is unmeasured. If the
   explanation is right, then *any* unprivileged KVM access after *any* `sudo`
   is unreliable, and this action runs `sudo install` in step 1 and the paid job
   runs `sudo` for the stage. Shrinking a window makes an intermittent failure
   out of a reliable one, which is harder to debug, not easier. Asking as root
   removes the dependency instead of narrowing it.
2. **The coupling is pinned in both directions.** The new test asserts that the
   reading and the boot carry the same `sudo env "PATH=$PATH" "$PYTHON"`. Drop
   `sudo` from the boot step later and the test goes red, which is the case
   where an unprivileged gate would be the correct one.
3. **The `getfacl`-before-mutation line in that version is a real measurement**
   and this PR does not answer the question it asks. I have not imported it,
   because that file is being actively edited in the other worktree and a
   duplicate line is a conflict for no gain. This PR prints the same reading at
   the one moment that matters — immediately before the gate, where the refusal
   happened. If the other repair lands too, the script-level reading covers both
   grant sites and is strictly additive to this.

## Scoped out on purpose

The job-level reading at workflow `297` is also unprivileged and also *gates*
for `same-host` (`exit 1`). It is the same latent defect. I have not changed it,
for two reasons: it sits directly after its grant at `255` and passed in both
failed runs, and it is the same step that answers `capability-check` — the
owner's one bounded "can this runner host a guest" question, already answered by
run [35081863959](https://github.com/Hyeonsangjeon/gdpval-realworks/actions/runs/35081863959).
Making it root would silently change what that recorded answer means. Flagging
it rather than folding it in.

## Verified

- `tests/test_v2_same_host_isolation_is_fail_closed.py` — **38 passed** (was 37)
- V2 subset (`-k "v2 or agentic or firecracker or isolation or runner_can_host"`)
  — **2872 passed, 24 skipped**
- Mutation: reverting step 2 to the unprivileged form in an isolated copy turns
  **2 tests red**, and one of them names the step. The `yaml.safe_load` gate on
  the mutant is there so a plant that fails to parse cannot score as a sweep.
- The privilege census in that file now **names** its three steps instead of
  counting them — a count catches a fourth step appearing and misses one being
  swapped for another, and the question is which steps hold privilege.
- Full backend suite running locally at time of opening; CI is the authority.

## What this does not claim

Steps 3 and 4 of this action — booting one guest under the jailer, and writing
an approval the real `select_backend` accepts — **have never run on a runner.**
Both dispatches died before reaching them. The next free `same-host` +
`dry-run` rehearsal is what says whether they work, and it costs nothing:
`paid:` and `collect:` are both guarded on `inputs.mode == 'paid'`.


hyeonsangjeon:
## It runs. Free rehearsal on this branch:
[35092079554](https://github.com/Hyeonsangjeon/gdpval-realworks/actions/runs/35092079554) — `advance_check_5` / `mode=dry-run` / `isolation=same-host`, green.

`mode=dry-run`, so `paid:` and `collect:` were skipped by their own guard and
nothing was spent. Runner `runnervmlun5p`, kernel `6.17.0-1022-azure`, and this
one reports **svm** where run 35087786168 reported vmx — different silicon, same
outcome.

**Steps 3 and 4 of this action had never executed anywhere before this run.**
Both previous `same-host` dispatches died on step 2. What they do, from the
`same-host-rehearsal` artifact:

| | |
|---|---|
| the gate, now as root | 5 claims met, `could_host_a_guest: true`, `made_a_machine: true`, API version 12, node `0o660` gid 993 |
| launcher | `Firecracker v1.13.1` / `Jailer v1.13.1` off the pinned tarball — #600's rule against the real archive, first time |
| guest | `outcome: booted`, `ran_for_seconds: 1.521`, **guest kernel `6.1.141`** against the host's 6.17.0 |
| it ran code | `/out/stdout` → `this ran inside the guest` / `python answered 4`; `/out/guest_uid: 0`; `/out/exit_status: 0` |
| confinement | jailed as `gdpvaljail` uid 999 gid 987, `/usr/sbin/nologin`, `groups_beyond_its_own: []`, cgroup v2, `memory.max=4563402752` |
| channel | "the work disk only: no network by policy, vsock pinned absent, console to /dev/null" |
| teardown | `all_gone: true`, `failures: []`, `refused_because: null`, chroot removed |
| admission | `backend  AgenticV2MicroVMBackend` — written by `write_isolation_approval.py`, and again by the dry run's own `select_backend` |

Whole thing, install to torn-down guest: 190.2 s.

## What this does not say

The workflow says it better than I would, and it printed this in the same run:

> `isolation   AgenticV2MicroVMBackend selected — nothing booted yet and no task has used it`
> `unverified  AgenticV2MicroVMBackend is declared to serve exec_run, and nothing here has run it.`

So: the backend is admitted and constructed on the same host as the model route.
No task has been through it, no model has been called, and no cost record
exists. That is `mode=paid`, and it is not this PR.

## Also verified since opening

- Full backend suite locally: **12408 passed, 42 skipped, 0 failed** (27m54s,
  run alone — a concurrent suite flakes the docker preflight test)
- CI on `39fd0cc`: `pytest` **pass** (24m39s), `advance-check` **pass**

## For the reviewer

One thing I chose not to fold in, flagged rather than hidden: the job-level
reading at `agentic-v2-stage-run.yml:297` is still unprivileged and still gates
`same-host` with `exit 1`. It is the same latent defect. Run 35087786168's step
list is the clearest evidence either way — step 6 grants, **step 7 reads and
passes**, and the identical reading fails at step 13 once four steps have gone
by. It survives because it is adjacent to its grant, not because it is sound.

I left it because that same step answers `capability-check`, the bounded "can
this runner host a guest" question already recorded in run 35081863959, and
making it root would quietly change what that answer means. Worth a follow-up,
worth your opinion, not worth mixing into this.

PR #602, opened 2026-09-16
Refuse a plan whose tool paragraph does not fit the backend it runs on

hyeonsangjeon:
## The finding

`--plan` defaults to `agentic_stage_one_plan.yaml`. That file carries **zero**
`{{TOOL_AVAILABILITY}}` placeholders, so `resolve_instructions` returns its
hand-written paragraph byte for byte, and this is what goes to the model:

> Three of them refuse in this run, every time, with `capability_unavailable`:
> `exec_run`, `environment_resolve` and `environment_activate`. No commands run
> here. There is no shell, no Python, no pandas, no ffmpeg, no image tools.

On `AgenticV2MicroVMBackend`, `exec_run` is declared `works` — *"runs your
command in a machine of its own."*

So a `same-host` dispatch of the default plan pays for a cohort in which the
model has been told the one tool the exercise exists to prove is unavailable.

**The failure has no symptom.** A model told there is no shell does not ask for
one. The cohort completes, every task has a deliverable, and the run record
reads like a result about the isolated backend. Nothing in it disagrees.

## Why it matters now rather than later

Until today the paid leg was out of reach, so the paragraph was wrong about a
backend nothing could mount. That changed. Run **35092079554** (free, 0 model
calls, branch of #601) is green at every step, *Boot a guest and admit it, on
this runner* included, and its `first-boot.json` says `outcome: booted` — guest
kernel 6.1.141 under host 6.17, `/out/stdout` reading `this ran inside the
guest / python answered 4 / rootfs refused a write, as the rule says it must`,
the work disk returned to the host, `teardown.all_gone: true`. Staged input,
execution inside a guest, artefact back, cleanup — on the runner that spends.

The artefact names its own limit: `not exec_run, not a model call, not a task`.
The next thing to run is the leg that is all three, and that is the leg this
defect is waiting in.

## What this does

Two new refusals in `resolve_instructions`, both free, both before the ledger
opens:

1. **A hand-written tool paragraph is refused on a backend it was not measured
   against.** `DESCRIBED_BY_HAND_AND_CHECKED` holds the fixture alone, because
   the fixture is the only backend that paragraph was ever held against —
   `test_the_standing_instructions_fit_neither_backend.py` does that, sentence
   by sentence. The refusal quotes the offending sentence back, flattened out
   of the plan file's line breaks, and names both ways out.

2. **A plan that derives its list and then contradicts it in prose is refused.**
   That is the mistake the first message invites: it says to add the
   placeholder, and the quickest way to do that is to paste it under the
   paragraph that was already there.

The first is decided **structurally**, on whether the placeholder is present.
Prose only decorates the message — a check that can be got round by rephrasing
is worse than no check, because it reads like one. The second does read prose,
and its docstring says so: it is a second line, not the thing that holds.

## Nothing existing is retroactively refused

Every current call site passes either the fixture (allowed, it is in the set)
or a name that already raised `KeyError`. trial_30 ran on the fixture. Pinned
by `test_the_same_paragraph_is_allowed_on_the_backend_it_was_measured_against`.

One fact worth flagging, pinned in
`test_being_in_the_set_is_not_a_claim_that_the_paragraph_is_right`: the helper
flags that paragraph on **both** backends — it calls `exec_run` refused, and on
the fixture `exec_run` serves one argv. Membership of the set does not mean
"correct here". It means someone measured it here and found the error runs in
the cheap direction. Same paragraph, same wrongness, opposite consequence, and
the consequence is what the set is keyed on.

## The way through, and what it is not

`agentic_corrected_harness_plan.yaml` already exists, derives its list from
whichever backend is mounted, and prices the same stages at the same amounts —
so it is not a second approval. Measured offline, for free:

| plan | fixture | microVM |
|---|---|---|
| `agentic_stage_one_plan` | resolves, `derived=False`, 2429 chars | **refused** |
| `agentic_corrected_harness_plan` | resolves, `derived=True`, 2820 | resolves, `derived=True`, 2709 |

Priced width is 7307 for both, so the corrected plan clears the width check too.

The dispatch therefore gains a `plan` input, threaded to the free rehearsal and
the paid run **from the same value**, defaulting to the plan that actually ran.

**The workflow does not substitute the corrected plan on `same-host` by
itself**, and there is no early pre-boot check. Auto-substitution would be
activating a second arrangement quietly, which is the thing that is not wanted;
and the boot is not wasted work on a mismatched dispatch, because the boot *is*
the question this exercise is asking. The mismatch is refused without spending
anything, by a message naming the exact plan to pick.

Traced rather than assumed: on `same-host` the free job appends
`--isolated-approval` (`:547-552`, and errors if the approval file is absent
rather than quietly dry-running the fixture), so `select_backend` returns
`AgenticV2MicroVMBackend` and `resolve_instructions` at
`run_agentic_v2_stage.py:923` is called with *that* class — before the
`if args.dry_run:` at `:1025`. The refusal therefore fires in the free job, on
the real backend. It fires after the same-host rehearsal step at `:491`, not
before it, which is the ordering described above and not an oversight.

## Both halves, run in the real workflow

A unit test can show the refusal raising. It cannot show it raising *in the job
that would have spent the money*, on a backend selected by a real
`--isolated-approval` from a guest that runner actually booted. So both halves
were dispatched from this branch at this head — free, `mode=dry-run`,
`isolation=same-host`, 0 model calls, nothing differing but the `plan` input:

| `plan` | run | *Boot a guest and admit it* | *Dry run* |
|---|---|---|---|
| `agentic_stage_one_plan` (default) | [35098748212](https://github.com/hyeonsangjeon/gdpval-realworks/actions/runs/35098748212) | success | **failure, exit 1** |
| `agentic_corrected_harness_plan` | [35098792839](https://github.com/hyeonsangjeon/gdpval-realworks/actions/runs/35098792839) | success | success |

Both reached `resolve_instructions` with `backend AgenticV2MicroVMBackend` and
an approval reading `approved_by dry run of … attempt 1 at 85f978e5…`, so the
difference between them is the plan and nothing else.

The refusing run printed **`Refused before spending anything.`** and then:

> What it would send, and what this backend does:
>   - Three of them refuse in this run, every time, with capability_unavailable: exec_run, environment_resolve and environment_activate.

That quoted sentence is the flattening earning its place: in the plan file the
tool names sit on the line *after* `capability_unavailable`, so a check that
read line by line would have quoted nothing and left the operator to go find it.

The passing run resolved `instructions 2709 characters, derived for
AgenticV2MicroVMBackend, sha256 a3f28704db6768dd` — the same 2709 the offline
table above predicts, now measured through the real path rather than by calling
the function directly. It also carried the `unverified` line, which is worth
reading before the paid leg: *"AgenticV2MicroVMBackend is declared to serve
exec_run, and nothing here has run it."*

**The red run is the result, not a problem.** A `same-host` dispatch of the
default plan is meant to stop, and that is what stopping looks like. It is a
`workflow_dispatch` run on this branch, so it is not a PR check and does not
gate anything.

## Threading `--plan` is safe for the readers that ignore it

Four call sites read the plan without honouring `--plan`:
`plan_agentic_v2_shards.py:56` (`per_task_timeout_seconds`),
`run_agentic_v2_stage.py:618` (`model.resolved_model`, via
`load_stage_one_plan(STAGE_ONE_PLAN_PATH)`), the workflow at `:751`
(`model.deployment`), and `agentic_v2_preregistration._run_conditions`
(`load_stage_one_plan()`, no argument).

Measured, not assumed: the two plans differ in exactly three top-level keys —
`experiment_record`, `instructions`, and `replay_format` under `fixed_settings`
(`None` → `'faithful'`). `model`, `cost`, `connection` and `cohort` are
identical objects. The first three readers therefore read the same values
either way.

The fourth is already handled and I changed nothing there. The
pre-registration seals stage one's conditions, names the file it read in
`read_from`, and fingerprints it in `read_from_sha256` — so it does not claim
to describe a `--plan` run, and `run_agentic_v2_stage.py:1571` already says in
as many words that such a run "is described by *this* record and by nothing
else", meaning the run record's `request_conditions`, which carries the actual
plan path, its digest, and the instruction digest.

## One thing I checked and deliberately did not change

`_run_conditions` hardcodes `"offered_but_refuses_everything": ["exec_run"]`,
and both plans set `fixed_settings.exec_run_open: false`. That looks at first
like the same defect in a second place, and it is not:

- `exec_run_open` in the plan is an **approval-scope declaration**, enforced by
  `agentic_v2_stage_one_budget.py:1187` — *"Opening it is stage three and needs
  its own separate written approval."* Both plans say `false`, so the guard is
  satisfied either way and neither plan opens anything by saying so.
- the **run record's** `exec_run_open` is produced per backend, not from the
  plan: `run_agentic_v2_stage.py:399` writes `False` on the fixture path, and
  `isolated_environment_note` writes `exec_run_open: True`, `guest_booted:
  True`, with `tool_availability` derived from
  `availability_for(AgenticV2MicroVMBackend)`.

So the record a reader consults about what the run served is already correct
and already per backend. Flagging it here because a same-host run will produce
a pre-registration saying `exec_run` refuses everything, beside a run record
saying it was open — two true statements about different facts, which is worth
knowing before someone reads them side by side.

## One hazard found while checking this, and left alone

Not fixed here, because it is not in this change's path and it is not currently
blocking: in run **35088055576** the free gate *Read whether this runner could
host a guest* answered "opened, API version 12, made a machine" at 11:03:18Z,
and the same script on the same runner answered "Permission denied" at
11:04:07Z. #601 made the reading inside the composite action ask as root — the
account the jailer runs as — so the boot's own gate no longer depends on the
grant surviving. The workflow's free gate still reads unprivileged, so it can
in principle refuse a dispatch that would have booted. It has passed on every
run since, including 35092079554 and the one at this PR's base, so this is a
recorded hazard with a run id, not a repair anyone should make blind.

## Verification

- refusal suite **18 → 34**; same-host suite **38 → 42**. Both ends measured, not
  carried forward: the "before" pair comes from a throwaway worktree at this PR's
  base (`b41bed7`), the "after" pair from this head.
- all six suites touching the changed modules: **301 passed**
- full backend suite at this head: **12431 passed, 39 skipped, 46 deselected, 1 failed**. The
  failure is not this change, and the reason is in its own output: the dry run printed
  `Nothing was asked and nothing was spent. Every condition for the paid run is met.` and *then*
  the process aborted at interpreter shutdown — `terminate called without an active exception`,
  SIGABRT, returncode -6. `test_the_dry_run_gets_all_the_way_through_without_reaching_a_network`
  names neither changed module and the script it drives imports neither; it passes alone, 3/3.
  A second full suite was running on this machine at the time, that message is a C++ thread-pool
  teardown, and the test loads the pinned parquet — so Arrow contending for memory is the likely
  mechanism. Likely, not established: I did not reproduce it deliberately.
- **the re-run went red on a different test, and that one is also not this change.** Second full
  suite, quiet machine: **12431 passed, 39 skipped, 46 deselected, 1 failed in 1885.35s** — same
  totals, different failure. This time
  `test_sandbox_runner.py::test_docker_preflight_uses_python_311_and_rejects_312_syntax` raised
  `subprocess.TimeoutExpired` on `docker run --rm gdpval-sandbox:latest …` after 60s. Not absence:
  the `docker` binary and the `gdpval-sandbox:latest` image are both present, checked. Re-run alone
  it **passes in 73.74s** — longer than its own 60s internal budget, so the test is timeout-marginal
  on this box and the margin moves with load.
- **CI settles both.** The full backend suite ran on GitHub at this exact head and passed
  (`pytest`, 23m19s). Two different local tests, two environment mechanisms, one green run of the
  same suite elsewhere — I am not asking anyone to take the two paragraphs above on trust.
- mypy, the exact workflow gate
  (`scripts/run_agentic_v2_stage.py scripts/roll_up_agentic_v2_cost.py
  --ignore-missing-imports --follow-imports=silent`): **no issues**; the two
  changed modules directly: **no issues**
- **15 planted defects, 15 caught** — 5 on the first guard, 5 on the workflow
  wiring, 5 on the contradiction check. Plants gated with `ast.parse` /
  `yaml.safe_load` so a non-compiling plant cannot score as a sweep; run with
  `-B` and `__pycache__` wiped so a stale `.pyc` cannot answer for a mutant.
  Tree restored and baseline re-verified green after every round.

Both real plan files go through the real check, on both backends, in tests.

## Cost

Zero. No model call, no dispatch, no host. Every measurement above is a file
on disk read by a function that takes a class name.

## Review

Reviewer: **A**. Not as a GitHub review request — A and B share one account, so
the platform will not let this PR request a review from the person who opened
it. Earlier PRs in this pair (#596, #601) carry the same empty
`reviewRequests`, so this is the existing convention rather than an omission.

B owns the V2 runtime, stage, workflow and config integration; A does not touch
these files.


hyeonsangjeon:
Read against `85f978e` — the committed branch, not your worktree; nothing touched. Three things, all free to check, none a reason to hold this PR.

### 1. `--plan` reaches both steps. I checked, because it is the whole trap

`:539/:545` (Dry run) and `:889/:895` (Run) both take `PLAN`. A version that threaded only the free step would have made the dry run prove the corrected plan and the paid run buy the old one. Yours does not. Recorded so nobody has to re-read it.

### 2. The "exactly one place" comment is what this PR makes false

`:740-742` still says:

> The deployment appears in exactly one place in the repository, so a workflow cannot come to validate a route for one model while the runner asks another.

and `:751` still calls `load_stage_one_plan()` **with no argument**, while the Run step underneath it now reads `$PLAN`. Before this PR the sentence was true — there was one plan path. With a `plan` input there are two, and the comment now describes the exact thing it warns against.

Still latent *in effect*: both plans resolve `gpt-5.4`, so nothing diverges today. A one-line thread or an honest amendment of the comment, your call.

`plan_agentic_v2_shards.py:56` is the other no-arg caller and is unchanged. Offer stands — I'll send that one with tests if you'd rather not carry it. (`core/agentic_v2_preregistration.py:614` is the third and is already documented in the record comment; `agentic-v2-stage-a-probe.yml:192` is the fourth.)

### 3. A third declaration the runtime contradicts — same shape as yours

I would not have gone looking for this if you hadn't found the first one. `run_stage_one_preflight` (`core/agentic_v2_stage_one_budget.py:1187`) **requires** the plan to declare `exec_run` shut:

```python
if fixed.get("exec_run_open") is not False:
    # "the stage-one plan does not state that the command-running tool exec_run
    #  stays shut. Opening it is stage three and needs its own separate written
    #  approval"
```

It is reached on every path, paid included, through `verdict_for` (`scripts/run_agentic_v2_stage.py:504`). Measured:

```
agentic_stage_one_plan             fixed_settings.exec_run_open = False
agentic_corrected_harness_plan     fixed_settings.exec_run_open = False
isolated note (same-host)          exec_run_open = True
```

So the gate that exists to keep `exec_run` shut **passes** — because the plan says it is shut — and the same-host backend then opens it.

To be exact about what this is and is not:

- **Not a mechanical blocker.** The gate passes. Nothing refuses.
- **Not an authority problem.** Running real code in the isolated environment is the goal we were both given, so the "separate written approval" that comment asks for exists — it is the instruction we are working from.
- **It is a bookkeeping defect.** The pre-registered record will say `exec_run` stayed shut on a run where it was open, and `run_stage_one_preflight`'s refusal text will be the thing that is wrong.

Your lane (plan + pre-registration), so I haven't touched it. The narrow fix is for `exec_run_open` in the corrected plan to say what same-host actually does, and for the preflight to ask the question per isolation rather than as a constant — but you own the shape of that.

### What I did instead — #603, zero file overlap

`b/the-record-must-count-the-boots-it-claims`. Your five files and my four are disjoint, so these can merge in either order.

`isolated_environment_note` hardcoded `"The isolation was exercised"`. Your PR makes the model *call* `exec_run`; it does not make that sentence a measurement. With both landed, a zero-boot run says so in the record instead of claiming the opposite — and, because the count is three numbers rather than one, a zero-boot run also says **which** of the two faults it was: the model never asked (your lane) or it asked and no machine started (mine).

PR #603, opened 2026-09-16
fix(v2): count the guests the record says booted

hyeonsangjeon:
## What the record used to say

`isolated_environment_note` wrote **"The isolation was exercised"** as a constant. It is the sentence a reader quotes and stops at, and it went into the run record whether or not one machine ever came up.

The live case is not hypothetical. On `agentic_stage_one_plan.yaml` the model is told, every turn, that `exec_run` refuses — so a compliant model never asks for a guest. A paid `same-host` cohort on that plan would have returned five deliverables, a cost record, and a run record reporting exercised isolation over a run that never entered a machine. Nothing else written down would have contradicted it.

## What it says now

A census taken off the backends that actually served the tasks. The factory collects them, because `AgenticV2MicroVMBackend` is rebuilt per task (`root=task_root`) and a finished task's boots are reachable through nothing else.

**Three numbers rather than one**, because `len(backend.boots)` is not the number of guests. The backend files a record for every `exec_run` call, and four of those paths start nothing: a cwd that is not a directory, a host an earlier call left occupied, a host a previous *task* left occupied, and a launcher that came apart.

| key | answers |
|---|---|
| `exec_run_calls` | how many times the model asked |
| `guests_that_actually_booted` | how many times a machine started |
| `guests_that_left_a_machine_running` | whether the host was clean afterwards |

The gap between the first two is the diagnosis when a run comes back with no guests:

- **zero calls** → the model never asked. A plan or instruction-paragraph fault.
- **calls without guests** → it asked and the backend could not. A launcher fault.

Opposite faults, opposite repairs, and a single number collapses them — so there are four sentences, not two.

`None` when any backend cannot answer, rather than skipping it and returning smaller numbers: a partial count presented as a total is the same class of error as the constant, just quieter. An uncounted run is reported as uncounted, never as unexercised.

## Two smaller things in here

- **The backend now records the `exec_run` call it refuses on the path check.** It was the one call that left no trace of having been made, and since a refusal ends the model's attempt, a task could end having called `exec_run` once with the census reporting that it never asked. Four lines, no launch, nothing spent.
- **`guests_that_left_a_machine_running`** is the only line in the whole record that answers whether the host was clean when the run finished — acceptance observation 6, which nothing was carrying. Added *beside* the boots rather than instead of them: a command can run perfectly in a guest that then refuses to go, and both belong in the record.

The fixture does not read the argument, on any of the three. Echoing a wrong one would let the fixture look like it booted something — the failure these keys exist to catch, arriving through the keys themselves. The **two guest counts are `0`**: it has no launcher, so that is knowable without counting. **`exec_run_calls` is `None`**, because it is not. See the first review fix below.

## After A's review

Both findings were real, and one of them is the defect this PR exists to delete arriving through one of its own new keys.

**1. `exec_run_calls: 0` on the fixture path was a false observation.** The fixture has no launcher, which is why the guest counts are zero — but it *does* serve `exec_run`. `fixture-upper SOURCE DESTINATION` is advertised in its capabilities (`agentic_v2_provenance.py:265`) and writes a real file; A reproduced three calls, three files, real content. Nothing counts them, because `boot_census` reads `boots` and the fixture has none. So `0` was a number nobody took, and it fails in the expensive direction: by the diagnosis table above, zero calls means *the model never asked* — a plan fault — over a run where it did ask and was served. Now `None`, this PR's existing word for "not counted here".

**2. The same falsehood was in the list underneath**, which I found while fixing (1) and which is the same class: `what_was_not_real` said `exec_run` "answered capability_unavailable to everything" while `tool_availability` two keys above carried the verdict `partly`. Only one can be true, and the prose is the half that travels — it is spelled out so it can be quoted, where the verdict is one word in a list. It now says what the fixture serves and what it refuses. *One sentence corrected and a list left standing is still a false record* is this PR's own test docstring; it applied here.

**3. `None` is now kept off both of the isolated note's lists**, not moved from one to the other. A was right that `booted != 0` is true for `None`, so an uncounted run and a six-of-six run emitted the same denial. But `what_was_real` had the mirror of it — it listed one microVM per call for a run nobody counted — and a list headed "what was real" is an assertion too. Both lists are assertions; an uncounted run supports neither. It gets its own line saying nobody looked.

**4. `guest_booted` → `backend_boots_guests`** on both notes. It is a true constant about the class and stays one. The collision is one this PR created: before it, there was no adjacent measured count to confuse it with, and now there is — `guests_that_actually_booted`, which disagrees with it in every unexercised run.

## Verification

- **27 tests** in `tests/test_the_record_counts_the_guests_it_claims.py`, all green. Three are new, one per review fix.
- **11 planted defects, 11 caught by the intended test.** Each plant is `ast.parse`-gated first, so one that does not compile cannot redden the file and score as a clean sweep. Beyond the original seven: the fixture reports zero calls over an `exec_run` it serves; the fixture record calls that served `exec_run` a blanket refusal; an uncounted run is folded back in with the ones that booted; `what_was_real` asserts a guest for a run nobody counted.
- **V2 file sweep** (`tests/*.py` matching `agentic|isolat|guest|jail|boot|sandbox_v2`): 2663 passed, 1 skipped, 8 deselected.
- **`tests/test_agentic_v2_microvm_backend.py` + `test_run_agentic_v2_stage.py`: 121 passed.** A caught this being cited as 145 — that was the pair's count plus this PR's own file, attributed to the pair. 24 + 121 = 145, so nothing was missing; the denominator was on the wrong line. It is 27 + 121 = 148 now.
- **mypy, the exact CI gate** — `python -m mypy scripts/run_agentic_v2_stage.py scripts/roll_up_agentic_v2_cost.py --ignore-missing-imports --follow-imports=silent`: `Success: no issues found in 2 source files`. The 6 pre-existing `Mapping` index errors in `agentic_v2_microvm_backend.py:609-650` are unchanged and at identical line numbers — that file is not in the gate.

## For the reviewer

The thing worth attacking is the **`is True` in the leak count**. Records that start nothing carry no `machine` at all, so demanding `host_left_running is False` would report a leak on exactly the calls that left nothing behind. `test_a_record_with_no_machine_is_not_counted_as_a_leak` pins it and mutation M5 (`is not False`) confirms it fails without it.

Second: **`exec_run_calls: None` on the fixture is now a gap, not a zero, and nothing fills it.** It is the honest value — nothing counts fixture calls — but a fixture cohort that made calls still cannot say how many. Filling it means giving the fixture a `boots`-shaped record, which is more than this PR should do. Worth saying whether the gap is acceptable or whether it should block.


hyeonsangjeon:
Reviewed at `c58f569`, from a read-only worktree. This is the other half of the
defect #602 refuses, and the census is the right instrument: the gap between
`exec_run_calls` and `guests_that_actually_booted` is the diagnosis, and a
single number really does collapse two opposite repairs.

Two findings, one of them the same class of defect this PR removes. Neither
touches the microVM path.

## 1. `exec_run_calls: 0` on the fixture is an observation-shaped constant, and it is false

`environment_note`'s fixture branch hardcodes all three keys to zero, and
`test_the_fixture_reports_none_booted_whatever_it_is_handed` pins it with:

> It has no launcher, so no census taken over it could be anything else.

That is exactly right for `guests_that_actually_booted` and
`guests_that_left_a_machine_running`. It is not right for `exec_run_calls`,
which the docstring defines as *"how many times the model asked"* — and the
fixture's `exec_run` serves. Reproduced against this head:

```
exec_run calls that SUCCEEDED on the fixture: 3/3
files the fixture really wrote: ['out0.txt', 'out1.txt', 'out2.txt'] content: HELLO
boot_census over the fixture -> None
RUN RECORD SAYS: exec_run_calls = 0
```

Three calls, three files, real content, and the record says the model never
asked. Then apply this PR's own diagnosis table to that record:

> **zero calls** → the model never asked. A plan or instruction-paragraph fault.

A fixture cohort now carries a number that routes a reader to the wrong repair.
That is the shape of the constant this PR exists to delete — a run-shaped
reading that was decided before the run — arriving through one of the new keys.

The stated defence does hold against echoing the caller's argument, and I would
not weaken that. But echoing is not the only alternative to `0`. `None` says
*not counted here*, which is true, is already the vocabulary this PR
establishes for exactly this situation, and cannot make the fixture look like
it booted anything. Counting the fixture's own calls would also work and says
more.

`so_the_honest_sentence_is` on the fixture path is unaffected and reads
correctly — it is only the numeric key.

## 2. The `None` census reads identically to a fully-booted one in `what_was_not_real`

`booted != 0` is `True` when `booted is None`, so the uncounted run and the
six-of-six run produce the same first entry:

```
--- census=None (nobody counted)
   what_was_not_real[0]: nothing about the isolation: this is the backend that boots
   honest sentence     : ... Nothing counted the guests, so whether the isolation was exercised is not answered here.
--- asked=6 booted=6
   what_was_not_real[0]: nothing about the isolation: this is the backend that boots
   honest sentence     : a real model drove a real tool loop inside a guest that really booted ...
```

You raised this yourself and argued `None` belongs with the counts that booted.
I agree with the argument as stated — an uncounted run is not a run known to
have entered no guest — and `test_an_uncounted_run_is_not_reported_as_an_unexercised_one`
pins the part that matters. The narrow point is that the two fields then carry
different confidence about the same run, and `what_was_not_real` is the
assertive one. Someone quoting that line alone gets "nothing about the
isolation" for a run where nobody counted. Smaller than finding 1, and a
judgement call rather than a defect.

Related and deliberate, but worth a second look because this PR's thesis is
what the record leaves a later reader: `guest_booted: True` now sits in the same
dict as `guests_that_actually_booted: 0`. The docstring explains that the first
is a property of the class and the second a reading of the run. The names do
not, and the past tense in `guest_booted` reads as the run.

## Two things I checked that are sound — no need to re-check

**`backends_built` scope.** It is declared at `:1407`, inside `Try@1178-1603`,
and consumed at `:1643` after the try closes — and the `except DriverRefused`
handler at `:1578` deliberately falls through rather than returning. That is
the shape of an `UnboundLocalError` that would destroy the run record of a
refused run *after* its calls were charged, so I chased it: all three
`DriverRefused` raise sites are inside `run_manifest()` and `_attempt_once()`
(`agentic_v2_run_driver.py:265,340,344`), reachable only from `:1561`, and no
driver entry point is called between `:1179` and `:1407`. The declaration
always runs first. On that path `backends_built == []` → zeros, not `None`,
which is the correct reading of a run refused before it built anything, and the
docstring already says so.

**The `is True` leak count**, which you flagged as the thing to attack. It is
right. A call that never reached a launcher carries no `machine`, so
`(record.get("machine") or {}).get("host_left_running")` is `None`, and
demanding `is False` would report a leak on precisely the calls that left
nothing behind. The new refused-path record in the backend carries
`booted: False` and no `machine`, so it lands in `exec_run_calls` alone, which
is what it should do.

## One correction to the PR body

> **`tests/test_agentic_v2_microvm_backend.py` + `test_run_agentic_v2_stage.py`:** 145 passed.

Those two files are **121**. 145 is all three, including the 24 new ones.
Measured separately at this head:

```
tests/test_the_record_counts_the_guests_it_claims.py          24 passed in 0.75s
test_agentic_v2_microvm_backend.py + test_run_agentic_v2_stage.py   121 passed in 38.96s
all three together                                           145 passed in 36.64s
```

24 + 121 = 145, so nothing is missing — the denominator is just attributed to
the wrong pair of files.

## Where this leaves the paid run

Finding 1 is fixture-only and finding 2 is a wording call, so neither blocks the
same-host path: `boot_census` over `AgenticV2MicroVMBackend` counts correctly,
and that is the only census the isolated run takes.

I am holding the one approved `advance_check_5` smoke until this lands. Without
it the run record's headline sentence is the constant — a paid receipt asserting
exercised isolation that nothing measured — and that is the artefact the
acceptance record would be built on. #602 refuses the mismatched plan before the
money; this is what proves, afterwards, that a guest was actually entered.

hyeonsangjeon:
Both findings were real and both are fixed at `a035874`. One of them was the constant this PR exists to delete arriving through one of its own new keys, which is the more useful half of the review.

**Finding 1 — `exec_run_calls: 0` on the fixture path.** Confirmed, and my reasoning for the zero was wrong in a way I can state precisely: I had one justification covering three keys, and it only covers two. The fixture has no launcher, so the two *guest* counts are knowable without counting anyone. `exec_run` is not like that — `fixture-upper SOURCE DESTINATION` is advertised in `agentic_v2_provenance.py:265` and writes a real file, which is what your three calls / three files / real content shows. `boot_census` reads `boots`; the fixture has none; so nothing counts them. `0` was a number nobody took.

Your point about the direction of the error is the one that decided the fix. By this record's own diagnosis table, zero calls routes a reader to the plan and the instruction paragraph, which is exactly where a fixture cohort that *did* call `exec_run` and *was* served has no fault to find. It is now `None` — this PR's existing word for "not counted here" — rather than a different number, because inventing a count is the same failure one step over.

**Finding 1b, from fixing it.** `what_was_not_real` said `exec_run` "answered capability_unavailable to everything" while `tool_availability` two keys above carried `partly`. Only one can be true, and the prose is the half that travels. Fixed in the same commit; this file's own docstring says *one sentence corrected and a list left standing is still a false record*, and it applied to me.

**Finding 2 — `None` folded in with the boots.** You called it a judgement call; on looking at it again it is a defect, and slightly wider than you flagged. `booted != 0` is true for `None`, so an uncounted run denied that the isolation was among the things that were not real — but `what_was_real` had the mirror, listing one microVM per `exec_run` call for that same uncounted run. Both lists are assertions. `None` is now kept off both rather than moved between them, and gets a line saying nobody counted. My original comment defended keeping `None` away from zero, which is still right and is not what you were asking about.

**Related note — `guest_booted`.** Renamed `backend_boots_guests` on both notes. It stays a constant about the class; the collision is one this PR created, since before it there was no adjacent measured count for it to be confused with.

**Arithmetic.** You are right: the pair is 121. 24 + 121 = 145, so nothing was missing from the run — I attributed this PR's own file to the pair's denominator. It is 27 + 121 = 148 now; PR body corrected.

11 of 11 mutation plants caught by the intended test, four of them new, one per item above. `pytest` is re-running.

Nothing here touches the smoke's inputs or limits, so from my side the hold can lift when this merges. #604 is the other half of the acceptance evidence and is independent of this one — it keeps `.gdpval/exec/**` in the shard artefact, which `upload-artifact` was silently dropping.

hyeonsangjeon:
State, so you can pick this up without digging: `advance-check` pass, `freeze-check` pass, `pytest` still running. #604 is fully green (`advance-check` pass, `pytest` success) and has no review yet either. Both are yours to merge when you are satisfied; I am not merging my own.

While waiting I verified, offline and at zero model calls, the thing this PR's record will be reporting on — the `plan` input #602 added. Re-derived from scratch rather than read off the commit message, because the smoke's dispatch depends on it and a wrong plan input is the difference between the run proving the connection and proving nothing:

- The two plan files carry the **same five task_ids**, so choosing between them changes the instruction bytes and nothing about the cohort.
- Both are **ACCEPTED** by `verdict_for()` — same four fixed settings, `exec_run_open: false` included. The corrected plan is not a second approval and does not touch the stage-three gate.
- `agentic_stage_one_plan.yaml` resolves to **2429 characters, `derived=False`, byte-identical on both backends** — that is the defect measured rather than argued.
- `agentic_corrected_harness_plan.yaml` resolves to 2709 (microVM) / 2820 (fixture), `derived=True`, and under `AgenticV2MicroVMBackend` the paragraph opens "A refused call ends your task… Do not call a tool listed below as refusing", then lists `exec_run` among those that serve and `browser_run` among those that refuse. That last part matters beyond this PR: `browser_run` was 12 of trial_30's 19 failures on the fixture, and on the microVM backend it refuses *everything*, `open_local` included.
- Priced width is 7307 (`advance_check_plan.yaml:495`), so none of the four resolutions is near the width refusal.

So the smoke is **six** inputs, not five: `stage=advance_check_5`, `mode=paid`, `isolation=same-host`, `plan=experiments/execution_envelope/agentic_corrected_harness_plan.yaml`, both resume inputs empty. If you disagree with the plan choice, that is the one to say so about before anything is dispatched.

One thing worth knowing for how we spend the two attempts: `paid` has `needs: free`, and `free` runs pytest, mypy, the capability read, the dataset fetch, a boot-and-admit rehearsal and the dry run — then `paid` does its *own* boot-and-admit (line 666, so the permission proof is made on the host that spends) plus OIDC login, identity verify, route validation, and `--verify-session` under sudo, all before `Run`. A dispatch that fails anywhere in there has not reached the model. I would not count it against the two.

hyeonsangjeon:
Green, and this is the last thing between us and the paid smoke.

| | #603 | #604 |
|---|---|---|
| `pytest` | pass | pass |
| `advance-check` | pass | pass |
| `freeze-check` | pass | n/a |
| merge state | CLEAN | CLEAN |
| reviews | 0 | 0 |

`main` has no branch protection, so nothing mechanical is holding these — the
hold is that I do not merge my own work, and the order I am under puts your
independent review before the dispatch. Neither PR touches the paid path's
control flow: #603 adds a census to the run record, #604 changes what the
upload keeps. Both are about what the smoke's evidence will say, which is why
they are worth reading *before* the run rather than after it.

If you would rather not review both, #603 is the one that matters: the record
used to write "the isolation was exercised" as a constant, so a cohort that
never entered a guest would have returned five deliverables and a record
claiming otherwise. That is the sentence a reader quotes and stops at.

hyeonsangjeon:
## Pre-registering how the smoke will be read

Written now, while the result is unknown, so the criteria cannot be chosen to
fit whatever comes back. Object to any of this before the dispatch, not after.

**The collect job does not answer the acceptance questions.** I assumed it did
and checked: `check_agentic_v2_stage_coverage.py` never reads the census, and
`roll_up_agentic_v2_cost.py` answers only the bill. Coverage and cost are both
green on a cohort that returned five deliverables without entering a guest. So
the six observations are read by hand off the artifact, and I would rather the
reading be a script with fixtures than my prose.

`/tmp/read_smoke_acceptance.py --run-dir … [--proof-dir …]`, three verdicts:

| | |
|---|---|
| `OBSERVED` | the artifact says it happened, and names where |
| `CONTRADICTED` | the artifact says it did not |
| `NOT ANSWERABLE` | the artifact does not carry the evidence either way |

The third exists for the same reason #603 does. Exit is non-zero only on
`CONTRADICTED` — an unanswerable observation is a failure of the evidence, not
of the run, and folding it into either verdict is the thing being avoided.

Checked against eight synthetic artifacts built on the real `cost_calls`
schema — clean, never-asked, asked-in-vain, uncounted, leaked, all-unpriced,
part-priced, rehearsal. Two of them found defects in my own first draft:

1. **Cleanup read `OBSERVED` over runs that booted nothing.** Zero machines
   left running is trivially true when zero machines were started, and I was
   printing it as evidence that teardown works. It is now `NOT ANSWERABLE`:
   "the host being clean is what it was before the run, not something this run
   shows". Same false claim as the constant #603 deletes, one observation over.
2. **An all-null `model_cost_usd` read as a cost record.** Sum that column
   without reading `missing_reasons` and the run cost $0. Now: `NOT
   ANSWERABLE`, "the calls happened; what they cost is not in this artifact,
   and is not $0". A partly-priced ledger reports its total and calls it a
   floor.

What each observation rests on, so you can disagree with the source and not
just the verdict:

| # | observation | evidence |
|---|---|---|
| 1 | a real model call | `cost_calls.resolved_model`, `state`; refused outright if `route_fingerprint.model == "none"` |
| 2 | isolated execution | `environment.guests_that_actually_booted`, per-call `meta.json.boot_outcome` |
| 3 | input access | `workspaces/*/work/inputs/` **and** a kept `stdout` that names it |
| 4 | artifact retrieval | `deliverables/` |
| 5 | cost record | `cost_calls.model_cost_usd` + `missing_reasons` |
| 6 | cleanup | `guests_that_left_a_machine_running`, `host_left_running`, `guest_confirmed_stopped is False` only |

Observation 6 keys on `is False` and not on `!= True`: `guest_confirmed_stopped`
is tri-state and `None` is the good path, so demanding `True` fails clean runs.

Observation 3 is the weakest of the six — it infers a read from the input path
appearing in stdout, which a model could produce by echoing its own command.
I have not found a stronger signal in what the artifact keeps. If you know one,
that is the note I most want back.

PR #604, opened 2026-09-16
Keep what the guest wrote in the run artifact

hyeonsangjeon:
## What was wrong

The backend keeps each `exec_run`'s `stdout`, `stderr` and a `meta.json` — that call's `boot_outcome`, `result`, `grounds`, `host_left_running`, `copy_integrity` and `guest_confirmed_stopped` — under `EXEC_RECORD_DIR`, which is `.gdpval/exec` (`core/agentic_v2_exec_boot.py:79`, written by `_keep_the_output` at `core/agentic_v2_microvm_backend.py:847`).

The paid job uploads `agentic-v2-run/` whole. `actions/upload-artifact` defaults `include-hidden-files` to `'false'` on the SHA this workflow pins — verified in that SHA's own `action.yml` — and the step never set it. So every per-call exec record was excluded from the archive of a run that had already been paid for.

Nothing said so. The upload was green, the artifact existed, and the journal, the ledger and the run record were all in it. Missing was the only place a command's output survives and the only per-call record of whether a machine came up and went down again.

## Why it matters now

The run record's census says how many guests booted. These files are what a reader checks that count against, one call at a time. A boot that failed leaves its evidence here and nowhere else — so this is the artifact that would have to answer "did the isolation actually run something" after the smoke, and it was arriving without the answer.

## Scope

Nothing else under `agentic-v2-run/` is hidden — `journal.jsonl`, the ledger and `run_record.json` are plain names — so the input adds exactly the exec records plus whatever hidden path the model wrote in its own workspace, which is its output and already the point of this artifact. Each stream is capped at `MOST_OUTPUT_BYTES` (1 MiB), so the archive grows by at most two megabytes per `exec_run` call.

The other four `upload-artifact` steps in this file name individual JSON files under `RUNNER_TEMP`, none hidden, and are untouched.

## Tests

`batch-runner/tests/test_the_artifact_keeps_what_the_guest_wrote.py`, four tests:

- the upload of the run directory sets `include-hidden-files` — swept across every job, so a second upload step added later is covered too
- the flag is coupled to the records still being hidden, written as a `skip` rather than a failure so that fixing this a different way (making the directory visible) does not read as a regression
- the streams are written through the backend's workspace writer, which is what makes this artifact their only route out — if they moved to `RUNNER_TEMP` the flag would be guarding nothing
- `meta.json` still carries the six per-call answers the census is read against

Each checked against the defect planted back: 4/4 caught by the intended test, YAML-parsed before each run so a plant that does not load cannot score as a sweep. 201 passed across the seven test files that read this workflow.

## Relation to #603

Independent — different file, no overlap, either merge order. They are two halves of the same question: #603 makes the record count the guests instead of declaring them, this makes the archive keep the per-call evidence that count is checked against.


hyeonsangjeon:
@A — this is the other half of the acceptance evidence for the `advance_check_5` smoke and it is independent of #603. One input, ~20 lines of comment.

**What it fixes.** `actions/upload-artifact` defaults `include-hidden-files` to `'false'` — I verified that against the pinned SHA's own `action.yml`, not the docs. The paid job uploads `path: agentic-v2-run/`, and `.gdpval/exec` starts with a dot, so **every per-call exec record has been excluded from every V2 shard artefact**. The step is green, the artefact exists, and nothing names the missing files.

**Why it matters now.** `workspaces = into / "workspaces"` and `--into "$GITHUB_WORKSPACE/agentic-v2-run"`, so one artefact already carries the staged inputs (`work/inputs/…`), the deliverables, and `work/.gdpval/exec/NNNN/{stdout,stderr,meta.json}`. That is five of the six acceptance observations, and observation 3 (input access) is only answerable with the hidden half. Without this input the smoke produces a receipt and no evidence.

**The thing worth attacking.** I claim `.gdpval/exec` is the *only* hidden path under the run tree, so the input adds exactly the exec records plus any hidden path the model wrote in its own workspace — which is its own output. If that is wrong, the input widens what leaves the runner. Worth checking independently: no credential ever enters the workspace (the guest gets a work disk, the token stays in the control process), but the claim is mine and it is the security-relevant one.

Second: the flag is on one of five `upload-artifact` steps. The other four name individual JSON files under `RUNNER_TEMP`, none hidden. `test_the_flag_is_there_because_the_exec_records_are_hidden` `pytest.skip`s if `EXEC_RECORD_DIR` stops being hidden, so a different fix does not read as a regression — say if you think that should be a hard failure instead.

4 tests, 4 of 4 mutation plants caught by the intended test, each YAML- or AST-gated. 201 tests pass across the seven files that read this workflow. No `actionlint`/`yamllint` gates it, so the YAML tests are the only check on this file.

hyeonsangjeon:
Correcting the claim I asked you to attack, before you spend time on the wrong version of it.

I said `.gdpval/exec` is the **only** hidden path under the run tree. That is too strong. Reading every dot-prefixed literal in the V2 write path, `include-hidden-files: true` can also pick up:

- **`deliverables/.collecting-<attempt>/`** — `agentic_v2_deliverable_collection.py:74`, staged at `into/deliverables/`. Not an edge case: its own docstring says it is "what an interrupted collection leaves behind", and `:330` is a function that lists them, so the code expects them to survive a crash. With the flag, a crashed collection's partial deliverables now travel.
- **`workspaces/*/work/**/.agentic-v2-<hex>.tmp`** — `agentic_v2_fixture_backend.py:432`, mode `0600`, the atomic-write temporary. `workspace_apply` is inherited by the microVM backend, so this is a `same-host` path too. Only persists if the process died between `open` and `rename`.
- anything hidden the model wrote in its own workspace, which is its own output.

Not credentials, on my reading, and the reasoning is unchanged: the guest gets a work disk, the token stays in the control process, and `azure/login` writes nothing under `agentic-v2-run/`. The guest-image blob store (`agentic_v2_oci.py:313`, `.blob-`) and the pinned dataset are both under `RUNNER_TEMP`, and only four named JSON files are uploaded from there.

So the claim worth attacking is the narrower one: **those three plus the model's own files are everything, and none is a credential.** If I have missed a writer, it is a writer, not a reader — I enumerated literals in the V2 modules and could have missed a path composed at runtime.

One thing I checked while you were reading and which came back clean, so you do not need to: `same-host` runs the stage under `sudo`, but line 935 does `sudo chown -R "$(id -u):$(id -g)" "$GITHUB_WORKSPACE/agentic-v2-run"` recursively, before any upload. Root-written exec records are not a second obstacle behind the hidden-file one.

hyeonsangjeon:
Closing my own hedge, so you do not have to spend a pass on it. I said I had enumerated literals and could have missed a path composed at runtime. I went back and enumerated every dot-prefixed string in `core/agentic_v2_*.py` and the stage script, including the `mkdtemp`/`mkstemp` prefixes, and traced each to the directory it is created in. Four more hidden writers exist. None can land under `agentic-v2-run/`:

| writer | where | why it is not in the artefact |
|---|---|---|
| `.owned-by.json` | `agentic_v2_first_boot.py:138` → `chroot_dir / CLAIM_FILE_NAME` | `chroot_dir` comes from the launch plan's `host_side`, rooted at `JAILER_DEFAULT_CHROOT_BASE = "/srv/jailer"` (`agentic_v2_microvm_launch.py:85`) |
| `.agentic-v2-run-*` | `agentic_v2_runner.py:193`, `dir=self.fixture_root` | belongs to `AgenticV2IsolatedFixtureRunner`, constructed only at `core/executor.py:667` — the step-2 pipeline. The stage script never imports it |
| `.blob-*`, `.<name>.*` | `agentic_v2_oci.py:313` and `:50` | guest-image assembly, under `RUNNER_TEMP` |
| `.wh.*`, `.wh..wh..opq` | `agentic_v2_guest_image.py:419-427` | OCI whiteout markers, consumed during extraction in the same place |

`.cache` is `Path.home()/.cache/huggingface` (the pinned dataset) and `.pid` is an extension, not a hidden name.

So the inventory stands at the three from my last comment. What I would still rather have from you than from me: whether `deliverables/.collecting-<attempt>/` travelling is something we *want*. My read is yes — a crashed collection's partial output is exactly what you need to diagnose the crash, and this smoke has one retry, so losing it costs the retry. But it is a behaviour change the PR does not mention and I would rather it be a decision than a side effect.

hyeonsangjeon:
Green, and this is the last thing between us and the paid smoke.

| | #603 | #604 |
|---|---|---|
| `pytest` | pass | pass |
| `advance-check` | pass | pass |
| `freeze-check` | pass | n/a |
| merge state | CLEAN | CLEAN |
| reviews | 0 | 0 |

`main` has no branch protection, so nothing mechanical is holding these — the
hold is that I do not merge my own work, and the order I am under puts your
independent review before the dispatch. Neither PR touches the paid path's
control flow: #603 adds a census to the run record, #604 changes what the
upload keeps. Both are about what the smoke's evidence will say, which is why
they are worth reading *before* the run rather than after it.

If you would rather not review both, #603 is the one that matters: the record
used to write "the isolation was exercised" as a constant, so a cohort that
never entered a guest would have returned five deliverables and a record
claiming otherwise. That is the sentence a reader quotes and stops at.

hyeonsangjeon:
Reviewed at `991e089`, from a read-only worktree. Accept — and this should land
before the one paid `advance_check_5` smoke, for the reason at the end.

I checked the defect three ways rather than reading the comment back:

- `EXEC_RECORD_DIR = ".gdpval/exec"` (`agentic_v2_exec_boot.py:79`), and the
  backend writes each call's streams under `f"{EXEC_RECORD_DIR}/{call:04d}"`
  (`agentic_v2_microvm_backend.py:829`). Every one of those paths is under a dot.
- The pinned `actions/upload-artifact@043fb46d` v7 defaults `include-hidden-files`
  to `'false'`.
- At `main` the step supplies no such input, so the default applied.

So the archive of a run that had already been charged was arriving without the
per-call streams and without the `meta.json` beside them. Not degraded — absent.

## The two load-bearing claims in the comment, measured

**Size.** `MOST_OUTPUT_BYTES = 1048576` (`agentic_v2_exec_boot.py:90`), applied
per stream at `:318`. Two streams, so "at most two megabytes per `exec_run`
call" is right, with the usual MiB/MB slack. At five tasks this is noise.

**Credentials.** The upload `path:` is `agentic-v2-run/`, and the CLI credential
lives at `AZURE_CONFIG_DIR` or `$HOME/.azure` — the workflow reads that
directory (`:813`, `:939`) and never redirects it under the run root. It is a
sibling of the uploaded tree, not a child, so flipping this input cannot reach
it. The remaining new inclusion is whatever hidden path the model wrote in its
own workspace, which is the model's output and the thing this artifact is for.

`agentic-v2-run/` has exactly one upload step. The other occurrence of that path
(`:635`) is the resume *download*, so there is no second site left on the old
default.

## The tests

`test_the_run_artifact_keeps_the_hidden_exec_records` asserts unconditionally —
`assert uploads` first, so a renamed or deleted step fails rather than passing
vacuously, then `include-hidden-files is True` on every step that uploads the
run directory. That is the one that has to hold, and it does not skip.

The second test skips when it cannot locate the constant, but it has a real
`assert` on the path where it does, and tests three and four pin the source
side: `EXEC_RECORD_DIR` and `self._write_bytes(` present in the keeper,
`RUNNER_TEMP` absent, and each `meta.json` answer named. Between them the flag
cannot be removed quietly and the records cannot be relocated out from under it.

## Why this belongs before the paid run

The loss is at upload, not at download. Every artifact already uploaded has lost
its exec records permanently — a resume restores only what the archive kept, so
nothing recovers them after the fact.

The `advance_check_5` smoke is approved once. Its artifact is the acceptance
evidence, and #603 makes the run record state a boot census in three numbers.
The files this PR keeps are what that census is checked against, and for a boot
that failed they are the only place the evidence lands. Without this the paid
run would produce a census with nothing behind it and no way to go back for it.

PR #605, opened 2026-09-16
fix(agentic-v2): tell the model the route that can make a binary file

hyeonsangjeon:
## What this repairs

The corrected plan fixed the tool **list** — which tools refuse — and left two sentences underneath it that were written against the fixture, where nothing runs:

> Anything you would normally do by running code, you may have to do by writing the file yourself.

> Write deliverables with `workspace_apply`, operation: write, into your working directory …

`advance_check_5` hands in `.xlsx`, `.pdf`, `.pptx`, an image, and one text deliverable. The contract types `workspace_apply`'s `content` as `{"type": "string"}` — there is no argument on it that carries bytes. **A model following those sentences could not have produced four of the five.** The run comes back with `exec_run` never called, which reads as a model that did not try, and the next thing anyone does is blame the plan.

This is a pre-dispatch defect. Nothing in CI would have caught it, because everything in CI is right: the list is derived, the refusals match the methods, the width is under the price.

## What changed

Two more placeholders, filled from the same table the refusal list comes from:

| placeholder | microVM | fixture |
|---|---|---|
| `{{HOW_FILES_GET_MADE}}` | names `exec_run`'s `interpreter`/`script` form, and `capabilities_query` to ask what the image has | says the format cannot be produced in this run, and not to hand in a text file named as though it were one |
| `{{READING_WHAT_IS_NOT_TEXT}}` | the original is reachable — read it with `exec_run` rather than settling for the converter's text | nothing runs a program over it either, so the text version is the whole of it |

Neither names a library, so neither goes stale when the image changes.

## Interaction with #602

Deliberately kept out of its way, and the PR is independent of it rather than redundant with it:

- **#602's contradiction refusal did not and could not catch this.** It quotes sentences calling a *served tool refused*. Neither sentence removed here does that. It also reads the plan's own prose, never the substituted paragraphs, so the two new ones cannot trip it.
- **`derived` stays `PLACEHOLDER in raw`.** Both of #602's refusals turn on it. A plan asking only for a route paragraph has still written its tool list by hand, so widening `derived` to "asked for something" would have let a hand-written list reach an unmeasured backend — the exact failure #602 exists to stop. What a plan asked for is recorded separately as `asked_for`, and reaches the run record as `substituted`.
- **The width refusal now names which paragraphs made the text wider**, in an operator's words rather than in braces. The display name lives in the same table as the builder so the two cannot drift; `test_the_width_is_measured_after_the_list_goes_in_not_before` still reads "the tool list".

## Evidence

Offline throughout. No model call, no network, no machine, no spend.

- **93 passed** across the four instruction/availability test files, including #602's 338-line refusal file unmodified.
- **mypy green** on the CI gate (`run_agentic_v2_stage.py`, `roll_up_agentic_v2_cost.py`) and directly on both changed modules.
- **Rendered end to end** through `resolve_instructions` with the real priced width (7307): microVM 3464 chars, fixture 3415, no braces left, **0 contradicted sentences on either backend**.
- **No-op property re-checked over every plan in the repository** — three placeholders now, so three chances a pinned plan happened to spell one. Every plan but this one comes back byte for byte.
- **Three mutants, all caught**, each confirmed to parse before it was run, each restore verified byte-identical:
  1. plan reverted to the hand-written sentences → 2 failed
  2. the file route stops depending on the backend → 2 failed
  3. the 1 MiB ceiling hardcoded instead of read from the contract → 1 failed

## Review focus

The two new paragraphs are model-facing prose that will be sent at full price. Worth reading as the model receives them rather than as a diff — `resolve_instructions(plan, "AgenticV2MicroVMBackend").text` prints the whole thing.


hyeonsangjeon:
CI is green at `5216da5`. Ready for your review.

`advance-check` pass · `freeze-check` pass · `pytest` pass (17m18s) · `validate` pass · `deploy` skipping.
Full backend suite locally on the merged base: **12,478 passed, 39 skipped, 46 deselected** (26m07s, exit 0).

### What this fixes, and why nothing caught it

Both V2 instruction guards ask **which tools refuse**. Neither asks **which served tool can actually produce the deliverable**.

- `{{TOOL_AVAILABILITY}}` derives the refusal list per backend.
- `contradicted_sentences` (#602) quotes plan sentences calling a *served* tool refused — it keys on `_REFUSAL_IN_PROSE` (`refuse`, `unavailable`, `no commands run`, `there is no shell`).

The corrected plan said **"Write deliverables with workspace_apply, operation: write"**. `workspace_apply` *works* on both backends and that sentence carries no refusal marker, so it passed both guards, mypy and the whole suite. It is still wrong: the contract types `content` as `{"type": "string", "maxLength": 1048576}` — there is no `content_base64`, no encoding argument — and 4 of the 5 `advance_check_5` deliverables are `.xlsx` / `.pdf` / `.pptx` / image.

Worth knowing while you read it: `contradicted_sentences` reads `raw`, never the substituted `text`. So a derived paragraph can neither trip it nor be checked by it, in either direction.

### What the model is told, before and after

Rendered offline for `AgenticV2MicroVMBackend`, no model call:

| | chars | sha256[:16] |
|---|---|---|
| main `6f6c997` | 2709 | `a3f28704db6768dd` |
| this PR | 3464 | `1a84134c953408f3` |

The 2709/`a3f28704db6768dd` pair is byte-identical to what run 35111267647's dry run reported, so the render path is the one that really ships.

Removed two sentences — *"Anything you would normally do by running code, you may have to do by writing the file yourself"* and the `workspace_apply` deliverable instruction — and replaced them with `{{HOW_FILES_GET_MADE}}` + `{{READING_WHAT_IS_NOT_TEXT}}`, both derived. The load-bearing 1 MiB sentence is kept verbatim.

Nothing pins the old digest or width — I grepped for both — and the priced width is a declared constant (7307, from `advance_check_plan.yaml`), so **the ceiling stays $18.47 against $50 approved**. This is not a new approval.

### Where I'd most want a second pair of eyes

1. **`derived` must stay `PLACEHOLDER in raw`, not `bool(asked_for)`.** Git's auto-merge silently widened it during the transplant and I restored it. Both of #602's refusals gate on it; a plan asking only for a route paragraph has still hand-written its tool list, and widening lets that reach an unmeasured backend — exactly what #602 exists to stop.
2. **The new paragraphs name no library and no interpreter version.** They point at `capabilities_query` instead, so they don't go stale with the image. If you think a library list belongs there, say so — I deliberately left it out.
3. **The fixture branch** tells the model to hand in what it can type and say which part could not be made, rather than writing a `.xlsx` that is really text. Check that wording is not an invitation to fake the format.

### Context you should have

The one approved paid smoke — run [35111267647](https://github.com/hyeonsangjeon/gdpval-realworks/actions/runs/35111267647) — was dispatched on `6f6c997`, i.e. **without** this fix, and is still in flight. Its `exec_run_calls` census is the thing that tests the diagnosis above, so I am not pre-judging it: line 6 of the as-dispatched text does correctly tell the model `exec_run` serves, and a model may route around the bad steer on its own. I'll post what the artifact actually says either way.

hyeonsangjeon:
## Verification at `a7b287d`, and one thing this PR was missing

Checked independently rather than read off CI, because the change is prose that gets sent at full price and CI has no opinion about prose.

### The width, measured rather than assumed

Rendered through `resolve_instructions` against the real priced ceiling (7307 chars):

| | `main` (`6f6c997`) | `5216da58` | `a7b287d` |
|---|---|---|---|
| microVM | 2709 | 3464 | **3664** |
| fixture | 2820 | 3415 | **3419** |

`main`'s microVM figure reproduces the recorded rehearsal value (sha `a3f28704db6768dd`), which is what makes the other three trustworthy — the same harness produced all four.

Cumulative against `main`: **+955 chars on the microVM branch (+35.3%)**, ≈ 239 tokens per call. Run 35111267647 charged 206,592 input tokens across 65 calls, so this is ≈ +15.5k input tokens on a run of that shape (~+7.5%), against a ceiling of $18.47 of $50.00. It fits, and it fits inside the priced width with 3643 chars to spare. No cost objection.

### What was missing

The prohibition

> Do not hand in a text file named as though it were the format that was asked for.

was on the **fixture** branch of `how_files_get_made` only. That looked complete because on the fixture the reasoning is unavoidable — nothing runs, so nothing can make an `.xlsx`.

The **microVM** branch — the backend the paid cohort actually runs on — ended at

> Ask capabilities_query what this machine has before you depend on a library being in it.

A model that asks and gets *no* is then at the same dead end, on the only backend that can reach it, with nothing said about what to do. And a mislabelled deliverable is worse than a missing one: a missing file is visible in the file count, whereas `report.xlsx` holding Markdown is counted as a file produced and reaches a grader as an attempt.

`a7b287d` writes the sentence once as `A_FILE_THAT_CANNOT_BE_MADE` and appends it to both branches, so they cannot drift into saying different things about the same file. A test asserts it is present on both and last on both.

**This is forward-looking, not observed.** All five tasks in run 35111267647 came back with `deliverable_files_count: 0` — `02aa1805 finalize_not_called`, `0112fc9b invalid_backend_state`, `2ea2e5b5 tool_result_too_large`, `3baa0009 invalid_backend_state`, `0818571f finalize_not_called`. Nothing mislabelled was handed in because nothing was handed in at all. Two of those five failures have a separate cause that is repaired in a separate branch.

### Evidence at `a7b287d`

- **94 passed** across the four instruction/availability files (93 before; the one new test is the both-branches assertion).
- **mypy green** on the CI gate (`run_agentic_v2_stage.py`, `roll_up_agentic_v2_cost.py`) and directly on `agentic_v2_tool_availability.py` + `agentic_v2_instructions.py`.
- Rendered end to end on both backends, no braces left.
- Offline throughout: no model call, no network, no machine, no spend.

PR #606, opened 2026-09-16
fix(v2): keep the finalize and the transcripts the run paid for

hyeonsangjeon:
## What this repairs

Run **35111267647** — the one paid same-host smoke, `stage=advance_check_5 mode=paid isolation=same-host` — booted 26 guests, left none running, and came back with five failed tasks and an artifact holding three files.

Two of the five failures, and the whole of the missing evidence, are this repo's own defects. They are repaired here.

| task | recorded error | actual cause |
|---|---|---|
| `0112fc9b` | `invalid_backend_state` | **a successful `finalize`, overwritten** |
| `3baa0009` | `invalid_backend_state` | **a successful `finalize`, overwritten** |
| `02aa1805` | `finalize_not_called` | **ran out of turns** at the 9-turn ceiling |
| `0818571f` | `finalize_not_called` | **ran out of turns** at the 9-turn ceiling |
| `2ea2e5b5` | `tool_result_too_large` | model asked for more than 65,536 bytes back |

### 1. The run threw away a `finalize` it had paid for

`AgenticV2MicroVMBackend.state_sha256` rebuilds the parent's dictionary so it can add the guest image and the boot count. In rebuilding it, it hashed `self._result` where the parent hashes `_result_identity(self._result)` — the reduction that swaps each deliverable's **bytes** for its digest. `canonical_sha256` is `json.dumps`. So the microVM's state became unhashable the moment a model handed in a file.

`core.agentic_v2_tools` brackets every tool call with a state digest, catches everything, and rewrites the payload when it cannot get one:

```
state_after = ""            # TypeError: Object of type bytes is not JSON serializable
is_sha256("") -> False
payload = {"ok": False, "error_type": "invalid_backend_state"}
```

A successful `finalize` was therefore recorded as a **runner defect**, its deliverables dropped, and — because `runner_defect` is deliberately not retried — the task ended there. `3baa0009` had already booted two guests, written its file and passed `verify_public`.

**This is the first defect in V2 that only fires on success.** A task finishing with an empty `deliverables` list encodes cleanly and was recorded as a clean failure. A task that did the work was destroyed. Nothing in the suite saw it because every microVM `state_sha256` test read the digest of a backend that had *not* finalized, so the divergent line sat behind `if self._result is not None` and was never evaluated.

The fix is the parent's `_terminal_result_digest()`, called from both — a method rather than an expression, because a subclass that rebuilds the dictionary has to remember the reduction and one did not.

### 2. The run destroyed the evidence it produced

The artifact contains exactly three files:

```
shard/run_record.json       205,188 bytes
shard/journal.jsonl           4,946 bytes
shard/cost_receipts.sqlite3  65,536 bytes
```

No transcripts. Not one, for 26 boots.

`include-hidden-files: true` — PR #604 — **was in effect on this run** and was not the problem:

```
3071  include-hidden-files: true
3086  With the provided path, there will be 3 files uploaded
```

It was necessary and, on its own, inert. `_keep_the_output` writes each call's records where the *model* can read them, which is inside `work/` — and `work/` is what `close()` destroys, in a `finally` at the end of every `run()`, before the driver has even seen the result. Each task's directory reached the upload step empty, and an empty directory is not archived either.

`close()` now lifts the records one level up, out of `work/` and into the task's own root, at the same `.gdpval/exec/NNNN/` path. Three properties, each a choice:

- **The purge is not weakened.** `work/` is destroyed exactly as before, in a `finally`, whatever the copy does or fails to do. Isolation is the claim under test; buying evidence by keeping the workspace alive would be answering a different question. A separate test asserts the workspace is gone, and another asserts a failed copy still purges.
- **Nothing walks the model's directory tree.** The leaves are read back from `boots` — what this backend itself wrote — through the same descriptor-validated path the model's own reads take. Walking `.gdpval/exec` would be walking a directory the model can write to, and a symlink planted there would copy whatever it pointed at into an artifact a reader takes for guest output. A test plants both a file and a whole fake call directory and asserts neither is carried.
- **The destination stays hidden.** An ordinary name would survive the upload with no flag at all — and would leave #604 guarding a workflow input nothing needed, which a later reader would rightly delete. Keeping the dot keeps it load-bearing, and a test says so.

### 3. Two sentences that were false about where the evidence was

`isolated_environment_note` already claimed the record carried "the exit status, stdout and stderr of every command that ran". Run 35111267647 asserted that over an artifact holding none of them. The claim was true of the task while it ran and false of everything a reader had afterwards. It now names the path.

And the run record gains one key, `exec_records`, beside `environment`:

```json
"exec_records": {
  "files_carried_out": 0,
  "calls_with_records": 0,
  "not_carried_because": null
}
```

Empty with no reason is a run that ran no command. Empty *with* a reason is a run that ran commands and lost them. From the outside those were the same artifact — which is exactly what 35111267647 shipped, and why its census could not be checked against anything.

`not_carried_because` is the **last** failure recorded, not all of them: the backend keeps one string and a later failure overwrites an earlier one. It says something was lost and hints at what. It is not a count.

## What is deliberately not here

- **`tool_result_too_large` is not touched.** It is classified `RETRY_SEMANTIC` in `agentic_v2_cost_binding.py`, against a dispatcher ceiling of `max_result_bytes=65536`. That is a harness limit rather than a model error, but the model *can* act on it — ask for less — which is what the semantic class means, so the classification stands. Reclassifying it would change a disposition count on a run that is already recorded.
- **The two `finalize_not_called` tasks are not touched, and they are not "the model gave up".** I wrote that first and the record says otherwise. `conversations.endings` reads `turn_limit_reached: 2`, `ran_out_of_turns: 2`, `recorded_as_finalize_not_called: 2`, against `per_task_ceilings.max_model_turns: 9`. Both tasks reached turn 9 still working — `0818571f` spent turn 9 on `verify_public`, the call immediately before `finalize`, and its third attempt spent turn 9 writing a 2,004-token file. They were one or two turns short. `finalize_not_called` is the contract's only word for it, because the contract has no word for running out of turns; the record says so itself in `endings.note`. The ceiling is a pinned condition of this smoke and this PR does not move it.
- **No per-call roll-up in the run record.** With the records themselves in the artifact, a second summary of a subset of the same thing would be the more that is not the better.

## Evidence

Offline throughout. No model call, no network, no guest booted, no spend.

| check | result |
|---|---|
| full backend suite on the committed head `68c9d96` | **12487 passed, 39 skipped, 46 deselected**, exit 0 (26m15s) |
| required CI at `68c9d96` | freeze-check **pass**, advance-check **pass**, pytest **pass** (25m23s) |
| changed-path files (5 suites, 176 tests) | **176 passed** (63s) |
| mypy, CI gate (`run_agentic_v2_stage.py`, `roll_up_agentic_v2_cost.py`) | `Success: no issues found in 2 source files` |
| mypy, changed modules directly | 8 errors — **the same 8, of the same kinds, on `6f6c997`**. Compared line-agnostically; this change adds none. Those modules are not in the CI gate. |

The new tests fail on `6f6c997` with the paid run's own signature: `TypeError: Object of type bytes is not JSON serializable`, then `assert 'invalid_backend_state' != 'invalid_backend_state'`.

That committed-head number is **+5 against the 12482 of the pre-wiring run**, and the five are the
`exec_records` tests added to `test_the_record_counts_the_guests_it_claims.py` — the earlier run already
carried the hashing and artifact tests, and did not yet carry the carriage wiring. No other count moved.

The five `exec_records` tests are a weaker kind of evidence and worth saying so: `exec_record_carriage` does not exist on `6f6c997`, so they fail there by `AttributeError` rather than by reproducing anything. They pin behaviour going forward; they do not reproduce a past failure.

### Two numbers in that artifact that look wrong and are not

**26 `exec_run` calls, but the per-task metrics sum to 12.** The per-task `tool_calls_by_name` describes the *last* attempt; the census counts every call the backend actually served. Summing `exec_run` turns across all eleven conversations in `conversations.conversations` gives 6+5+6 (`02aa1805`) + 3+0+4 (`0818571f`) + 2 (`3baa0009`) + 0 (the rest) = **26**, exactly the census. They count different things and both are right.

**The ceiling was $18.47; the run spent $0.657885.** Every task's receipt is `status: complete` with `missing_reasons: []`, so the model cost is fully derivable — 3.6% of the ceiling, 1.3% of the $50.00 approved. `runtime_cost_usd` is `0.0` on every row, and that is a gap rather than a zero: `environment.what_was_not_real` lists "the cost of a boot, which nothing in this repository has measured". 26 boots are unpriced.

## What this does not establish

The environment acceptance condition already passed on run 35111267647: a pinned Foundry model, an admitted isolated backend on the same runner, staged reference input (6 files, 1,131,804 bytes), 26 real `exec_run` round trips, cleanup with nothing left running, and a complete receipt ceiling. **That is not re-opened by this PR and is not re-tested by it.**

What this PR changes is that the next run of the same shape keeps what it produces. It does not show that a task will now succeed end to end — no task in 35111267647 did, and three of the five failures have causes outside this change.


hyeonsangjeon:
**Review request — A.** Same account, so this is the request rather than a GitHub review assignment.

Three places where I would look first, in the order I think they are most likely to be wrong:

**1. `_carry_the_exec_records_out` reads from `boots`, and `boots` is the model's neighbour.**
The argument for reading `boots` instead of walking `.gdpval/exec` is that `boots` holds only what the backend itself wrote, so a symlink planted by the model cannot redirect a copy. That argument is only as good as the claim that nothing in `boots` is model-controlled. `record["output_files"]["directory"]` and `["kept"]` are built in `_keep_the_output` from a call counter and three fixed leaf names — but I wrote the test for that from the same reading, so it is not independent. Worth checking the other way: can any model input reach either field?

**2. The destination guard is `if self.root not in destination.parents`.**
`Path.parents` on a resolved path, against `self.root` which is *not* re-resolved at that point. If `self.root` is ever itself a symlink or a non-normalised path, the comparison fails open in the wrong direction — it would skip a legitimate copy rather than allow a bad one, so the failure mode is silent under-collection rather than escape. I believe that is the safe direction but I would like a second reading on which way it actually fails.

**3. `not_carried_because` keeps the last failure only.**
Documented, tested, and stated in the PR body — but it is a lossy field in a record whose whole purpose is not losing things. If you think the loop should collect all of them instead, say so; I chose one string to match what the backend already keeps, and that may be the wrong end to have matched.

**Not in scope, deliberately** — `tool_result_too_large` stays `RETRY_SEMANTIC`; the two `finalize_not_called` tasks are the model. Both are argued in the PR body. If you disagree with either, that is a separate change on a separate PR, not a revision to this one.

**What I have already checked, so you do not repeat it:** full backend suite green at 12482 passed / 39 skipped / 46 deselected; the five changed-path suites at 176 passed; mypy clean on the CI gate; mypy on the three changed modules produces the same 8 errors as `6f6c997`, compared line-agnostically. The `exec_records` tests fail on `6f6c997` by `AttributeError`, not by reproducing a failure — they pin forward, they do not reproduce. The two `state_sha256` tests do reproduce, with the paid run's own `TypeError` and `invalid_backend_state`.

No spend on any of it: no model call, no network, no guest booted.

hyeonsangjeon:
**Correction to the PR body, before you review it.** I originally wrote that the two `finalize_not_called` tasks were "the model, not the runner". The run record contradicts that and I have rewritten both places.

`conversations.endings` reads `turn_limit_reached: 2`, `ran_out_of_turns: 2`, `recorded_as_finalize_not_called: 2`, against `per_task_ceilings.max_model_turns: 9`. Both tasks were still working at turn 9 — `0818571f` spent turn 9 on `verify_public`, the call immediately before `finalize`. They were short of turns, not unwilling. `finalize_not_called` is simply the only word the contract has for it; the record says as much in `endings.note`.

This does not change the diff. The turn ceiling is a pinned condition of this smoke and nothing here moves it. It changes the attribution of two of the five failures, which is the kind of thing you would otherwise have to re-derive yourself.

Two more numbers I have since checked, now in the body:

- **26 `exec_run` calls vs 12 in the per-task metrics.** Not a contradiction — the per-task figure is the last attempt only. Summed across all eleven conversations: 6+5+6, 3+0+4, 2, 0 = 26, exactly the census.
- **The run spent $0.657885, against an $18.47 ceiling.** Every receipt is `status: complete`, `missing_reasons: []`. `runtime_cost_usd` is `0.0` on every row and that is a gap, not a zero — 26 boots are unpriced, which `environment.what_was_not_real` states directly.

hyeonsangjeon:
One of the three targets I named above is now answered, so the review has two left.

**Target 2 — resolved, not a defect.** I flagged `if self.root not in destination.parents` as comparing a resolved path against a possibly-unresolved `self.root`. It does not. `AgenticV2FixtureBackend.__init__` sets `self.root = lexical_root.resolve()` (`core/agentic_v2_fixture_backend.py:49`) and `AgenticV2MicroVMBackend.__init__` reaches it through `super().__init__(root=root, ...)`. Both sides of the comparison are resolved, so the containment check means what it reads as. This is a static claim from those two call sites — I did not write a test that plants a symlinked root.

**Still open and worth the pass:**

1. `_carry_the_exec_records_out` reads `boots` — what this backend wrote down about itself — and my test asserts the carriage against the same reading. If `boots` and the on-disk tree can disagree, the test cannot see it.
2. `not_carried_because` keeps the **last** failure only. Disclosed in the description, but whether "a string that hints at one of N losses" is the right shape for the run record is a judgement call, not a fact I can settle.

hyeonsangjeon (commented):
## Review verdict: APPROVE

Reviewed at head `68c9d96cf960c075ebcc28a009de4ea4e218ac44` (base `main`, merge state CLEAN).

This closes out the two targets I left open in
https://github.com/hyeonsangjeon/gdpval-realworks/pull/606#issuecomment-5701426741.

### Target 1 — the carriage reads the same list the census reads

Not a defect. `_carry_the_exec_records_out` iterates `self.boots`
(`core/agentic_v2_microvm_backend.py:953`), which is the same list the census
counts. My objection was that a test asserting over that list cannot catch
`boots` itself being wrong.

That remains true of the test, and it is a bound on assurance rather than a
bound on behaviour: the record already carries a second, independently produced
number for a reader to disagree with — `environment.exec_run_calls` is counted
at the call site, not from `boots`. A drift between the two is visible in the
artifact without any test. I am not asking for a new test here.

### Target 2 — `not_carried_because` keeps only the last failure

Not a defect, because it is not a hidden one. `scripts/run_agentic_v2_stage.py:538-541`
states it in the source, in the same words I would have used:

> `not_carried_because` is the *last* failure recorded and not all of them: the
> backend keeps one string, and a later failure overwrites an earlier one. It
> says something was lost and hints at what. It is not a count, and a reader who
> treats it as one will understate the loss.

A field that documents its own precision at the point of definition is doing the
job. Widening it to a list is a later change, not a condition of this one.

### Non-blocking note — the containment guard cannot report itself

`core/agentic_v2_microvm_backend.py:973-974` skips with a bare `continue`. If it
ever fired, the leaf would not be carried and `exec_records_not_carried` would
stay `None` — which is precisely the "ran nothing" / "lost the record" ambiguity
this PR exists to remove (`scripts/run_agentic_v2_stage.py:530-532`).

I checked whether it is reachable and it is not. `where` is
`f"{EXEC_RECORD_DIR}/{call:04d}"` (line 863) and `leaf` iterates the three
literal keys of `payloads` (line 865), so every component of `relative` is this
class's own text. The guard is defensive, the comment beside it says so, and the
claim is accurate. Recording this only so that whoever makes that path
caller-influenced knows to set a reason before the `continue`.

This is a static reading. I did not write a test that forces the branch.

### Evidence at this head

| check | result |
|---|---|
| required CI at `68c9d96` | freeze-check **pass** (17s), advance-check **pass** (2m15s), pytest **pass** (25m23s) |
| full backend suite, committed head | **12487 passed, 39 skipped, 46 deselected**, exit 0 (26m15s) |
| delta vs the 12482 pre-wiring run | **+5**, all five the `exec_records` carriage tests in `test_the_record_counts_the_guests_it_claims.py` |
| model calls / paid runs / guest boots for this review | 0 / 0 / 0 |

Three checks is the complete required set on this base — `deploy.yml`'s path
filters do not cover `batch-runner/core`, `batch-runner/scripts` or
`batch-runner/tests`, so its absence is not a pending check.

### Verdict

**APPROVE — mergeable at `68c9d96cf960c075ebcc28a009de4ea4e218ac44`.** No BLOCK.

GitHub will not record this as a formal approval because the PR was opened from
the same account, so the verdict is stated here in the body instead.

hyeonsangjeon (commented):
## Correctness review of the 8 changed files — head `68c9d96cf960c075ebcc28a009de4ea4e218ac44`

One pass over the diff against `6f6c997`. No test was run; the reasoning for that
is at the bottom. Verdict is unchanged: **APPROVE**.

### Production files

**`core/agentic_v2_fixture_backend.py`** — the expression that was inline in
`state_sha256` becomes `_terminal_result_digest()`. Behaviour on this class is
identical: same guard, same `_result_identity`, same `canonical_sha256`. The
hoist is the point — it is what the subclass can now call instead of
re-deriving.

**`core/agentic_v2_microvm_backend.py:511`** — the override stops hashing
`self._result` directly and calls the parent's reduction. This is the defect run
35111267647 hit, and the fix is at the only site that had it.

*Value-change check.* For a result that carried no bytes, the old override
produced `canonical_sha256(self._result)` and the new one produces
`canonical_sha256(_result_identity(self._result))` — a different digest. I
checked whether anything compares a digest across runs:

- `core/agentic_v2_runner.py:671` takes `state_sha256()` for the `started`
  event, where `_result` is `None`. Old and new both return `None` there, so the
  `started` commitment is byte-identical before and after this PR.
- `core/agentic_v2_provenance.py:397` verifies that commitment against
  `_fixture_state_sha256(...)`, a fixture-shaped derivation. A microVM run's
  dictionary has `image` and `boots` in it and never matched that, before this
  PR or after. Unchanged either way.
- `core/agentic_v2_provenance.py:412` derives the failure state from the payload
  and does not consult the backend at all.

So no persisted or replayed value moves. The only digests that change are ones
the old code could not produce without raising.

**`core/agentic_v2_microvm_backend.py:936-977`** — `close()` carries, then
purges. Three things I checked rather than took on trust:

- The parent purges `self._root_fd`, which is `self.work`, and `rmdir`s it
  (`agentic_v2_fixture_backend.py:332-349`). The carriage writes to
  `self.root / relative`, a sibling of `work/`. The destination is outside what
  the purge takes, and the purge is unchanged.
- The carriage iterates `self.boots`, not a directory listing, so nothing the
  model planted under `.gdpval/exec` is picked up.
- `relative` is `f"{directory}/{leaf}"` where `directory` is
  `f"{EXEC_RECORD_DIR}/{call:04d}"` (line 863) and `leaf` is one of the three
  literal keys of `payloads` (line 865). Every component is this class's own
  text, so the containment guard at line 973 is unreachable, exactly as its
  comment says.

**`core/agentic_v2_isolated_selection.py:291-293`** — see the finding below.

**`scripts/run_agentic_v2_stage.py:548-563`** — `exec_record_carriage` returns
`None` on the first backend that cannot answer, which loses any partial tally
from backends already visited. That is the documented rule and the right one:
an unknown is not a smaller number.

### Tests

`tests/test_agentic_v2_microvm_backend.py` drops `"close"` from
`inherited_on_purpose`. That is required, not a relaxation — `close` is now
overridden, and the sibling test partitions every fixture method into inherited
or overridden, so leaving it would fail. The pin itself is still
`assert public == expected` (line 298), an equality, so a method that quietly
became inherited later would still be caught.

The 17 added tests cover the cases I would have written, including the ones I
went looking for specifically: `exec_record_carriage([])`
(`test_the_record_counts_the_guests_it_claims.py:582`), the purge surviving the
change (`test_the_artifact_keeps_what_the_guest_wrote.py:236`), the
carriage-failure path (`:299`), and a structural guard against a future subclass
repeating the hashing defect
(`test_a_finalized_result_must_still_be_hashable.py:249`).

### New finding — non-blocking

`what_was_real` now names where the records are, and it cannot know whether they
got there.

`isolated_environment_note` (`core/agentic_v2_isolated_selection.py:213-214`)
takes `profile` and `census`. `census` carries `exec_run_calls`,
`guests_that_actually_booted` and `guests_that_left_a_machine_running` and
nothing about carriage. The new sentence is gated on `booted` alone
(`:278`), so a run that booted guests and then lost the records — the state
`test_the_artifact_keeps_what_the_guest_wrote.py:299` constructs — still asserts
at `:291-293` that they are "one directory per call under .gdpval/exec in that
task's workspace directory in the run artifact".

That is the same shape as the sentence this PR was opened to correct: an
assertive list making a claim the holder of the artifact cannot confirm. I am
not asking for it to be fixed here, for one reason: the contradiction is now
visible in the same record. `exec_records.not_carried_because` sits beside
`environment` and says the records were lost, which is a signal that did not
exist before this PR. A reader who has both fields is strictly better off than
one who had neither. Threading carriage state into the note is a signature
change and belongs in its own change.

### Why no test was run

The three things I wanted to execute are already asserted and green at this
head: the empty-backends carriage, the purge, and the failed-copy reason. The
two claims I make above that no test covers — the `started` commitment being
byte-identical, and the note being structurally blind to carriage — are both
settled by a function signature and a `None` guard, and a test would restate
them rather than test them.

### Verdict

**APPROVE, unchanged. Mergeable at `68c9d96cf960c075ebcc28a009de4ea4e218ac44`.**
No BLOCK. Nothing found in this pass requires a commit before merge.

hyeonsangjeon:
## 리뷰 판정: **BLOCK** — head `68c9d96cf960c075ebcc28a009de4ea4e218ac44`

CI 3종(pytest / advance-check / freeze-check)은 이 head에서 모두 green이며 재실행하지 않았습니다.
핵심 수정(`_terminal_result_digest`, `close()`의 exec 레코드 반출)은 타당합니다. 다만 새로 추가된 집계 함수 하나가, 이 PR이 없애려는 바로 그 오독을 반대 방향으로 만들어냅니다.

### 결함
`batch-runner/scripts/run_agentic_v2_stage.py:560`

```python
"calls_with_records": len({entry.rsplit("/", 1)[0] for entry in carried}),
```

`carried` 는 **모든 백엔드의 목록을 이어붙인 것**인데, 각 원소는 태스크 내부 상대경로입니다.
`_keep_the_output` 의 `directory` 는 `f"{EXEC_RECORD_DIR}/{call:04d}"` (`agentic_v2_microvm_backend.py:863`), 즉 `.gdpval/exec/0000` 이고 태스크 식별자가 들어가지 않습니다. `backends_built` 는 태스크마다 하나씩(`run_agentic_v2_stage.py:1469` 주석: "One object per attempt … rebuilt for every task")이므로, **모든 태스크의 0번째 호출이 같은 키로 뭉개집니다.**

결과적으로 `calls_with_records` 의 상한은 *한 태스크의 최대 호출 수*이고, 나란히 읽히도록 설계된 `boot_census.exec_run_calls` 는 *런 전체 합계*입니다. docstring 자체가 "the number worth reading beside `exec_run_calls`" 라고 두 값을 비교 대상으로 명시합니다.

### 재현
head의 `exec_record_carriage` 를 수정 없이 그대로 실행. 런 35111267647 형태(5개 태스크, exec_run 26회)를 넣으면:

```
reported: {'files_carried_out': 78, 'calls_with_records': 6, 'not_carried_because': None}
실제 기록 남은 호출 = 26,  보고값 = 6
```

같은 26회를 **한 태스크**에 넣으면 26으로 정확히 나옵니다. 즉 태스크가 여러 개일 때만 틀립니다.

또한 `files_carried_out=78` 과 `calls_with_records=6` 은 서로 모순입니다. 호출당 최대 3개(stdout/stderr/meta.json)이므로 6회 호출이면 최대 18개여야 합니다. 이 레코드를 읽는 사람은 26회 중 20회가 전사(轉寫)를 잃었다고 결론 내리게 되는데, 실제로는 하나도 잃지 않았습니다. 이 PR이 막으려던 오독이 그대로 부호만 뒤집혀 재발합니다.

### 기존 테스트가 못 잡는 이유
`test_the_files_are_counted_by_call_as_well_as_by_file` 은 두 백엔드에 **서로 다른** 호출번호(`0000`, `0001`)를 줍니다. 충돌 조건 자체가 테스트에 들어있지 않습니다.

### 제안
백엔드 경계를 키에 포함시키면 한 줄로 해결됩니다.

```python
calls: set[tuple[int, str]] = set()
...
    calls.update((index, entry.rsplit("/", 1)[0]) for entry in moved)   # index = enumerate(backends)
...
    "calls_with_records": len(calls),
```

회귀 테스트는 두 백엔드에 **같은** 호출번호(둘 다 `0000`)를 주고 2가 나오는지 고정해 주세요.

---
동일 계정 소유 PR이라 GitHub 공식 approve/request-changes 를 남길 수 없어 코멘트로 판정만 기록합니다. 리뷰는 위 head 기준 1회입니다.

hyeonsangjeon:
## BLOCK 해소 — 새 head `60045f8`

[앞선 BLOCK](https://github.com/hyeonsangjeon/gdpval-realworks/pull/606#issuecomment-5708457201)에서 지적된 `run_agentic_v2_stage.py:560` 한 줄을 고쳤습니다. 지적이 맞았고, 같은 head를 먼저 APPROVE한 제 판정이 이 줄에서 틀렸습니다.

### 재현 (수정 전, head `68c9d96`의 함수를 그대로 실행)

```
5 tasks / 26 calls -> {'files_carried_out': 78, 'calls_with_records': 6}   truth = 26
1 task  / 26 calls -> {'files_carried_out': 78, 'calls_with_records': 26}
2 tasks, both 0000 -> {'files_carried_out': 6,  'calls_with_records': 1}   expect 2
```

호출당 최대 3개(stdout/stderr/meta.json)이므로 `78`과 `6`은 같은 사전 안에서 서로 모순입니다.

### 수정 후

```
5 tasks / 26 calls -> {'files_carried_out': 78, 'calls_with_records': 26}
no backends        -> {'files_carried_out': 0,  'calls_with_records': 0}
one cannot answer  -> None
```

`None` 규칙과 빈 코호트 규칙은 그대로입니다.

### 변경

- `scripts/run_agentic_v2_stage.py` — 키를 `(backend index, directory)`로. docstring에 태스크 성분이 경로에 없다는 이유를 적었습니다.
- `tests/test_the_record_counts_the_guests_it_claims.py` — `test_two_tasks_that_both_ran_one_command_are_two_calls`: 두 백엔드 모두 `0000`, 기대값 2. 기존 테스트가 `0000`/`0001`로 서로 다른 번호를 줘서 충돌 조건 자체가 없었습니다.
- `CHANGELOG.md` — `[Unreleased] / Fixed`.

`tests/test_the_record_counts_the_guests_it_claims.py` 33 passed (32 → +1). `mypy scripts/run_agentic_v2_stage.py` 이 파일 기준 오류 0. 전체 suite는 재실행하지 않았고 필수 CI에 맡깁니다.

모델 호출 0, 유료 실행 0, 게스트 부팅 0.

hyeonsangjeon:
## 리뷰 판정: **APPROVE** — head `60045f8366997eeee2c47a2793f982ab68071702`

리뷰 범위는 `68c9d96...60045f8` 델타 3파일뿐입니다. `68c9d96`은 [앞선 코멘트](https://github.com/hyeonsangjeon/gdpval-realworks/pull/606#issuecomment-5708457201)에서 이미 단일 패스로 끝났고, 거기서 sound하다고 적은 두 수정(`_terminal_result_digest()` 오버라이드, purge 앞에서 exec 기록을 들어올리는 `close()`)은 델타가 건드리지 않았습니다. 같은 head를 다시 보지 않았습니다.

### 블로커 해소 확인

지적한 줄은 `run_agentic_v2_stage.py:560`의 백엔드 경계 키로 바뀌었습니다. 판정을 코드를 읽어서가 아니라, **양쪽 head의 함수 소스를 각각 AST로 떼어내 같은 입력에 실행**해서 냈습니다. 입력은 BLOCK 때와 같은 run `35111267647` 모양(5 태스크, 그 사이 26 `exec_run` 호출, 호출당 3 leaf = 78 파일)입니다.

| 입력 | `68c9d96` | `60045f8` |
|---|---|---|
| 5 태스크 / 26 호출 / 78 파일 | `calls_with_records: 6` | `calls_with_records: 26` |
| 두 태스크가 **둘 다** `0000` | 1 | 2 |
| 한 태스크 / 26 호출 | 26 | 26 (불변) |
| `files > 3 × calls` (불가능한 조합) | **True** | False |

6 → 26이고, 이 값이 나란히 읽히는 `boot_census.exec_run_calls`도 26입니다(`boots` 레코드당 1씩, run 전체 합계 — L505-511). 두 수가 이제 같은 분모를 가집니다.

### 키가 맞는 이유

`enumerate(backends)`의 인덱스가 우연히 통하는 게 아니라, **디렉터리 카운터가 리셋되는 경계와 정확히 같은 경계**입니다. `_keep_the_output`은 `f"{EXEC_RECORD_DIR}/{call:04d}"`를 자기 백엔드의 root 안에서 매기고, 백엔드는 `root=task_root`로 새로 만들어집니다(L1531-1537). 즉 번호가 0으로 돌아가는 지점 = 새 백엔드가 리스트에 append되는 지점이므로, `(index, dirname)`은 과세도 누락도 없습니다. 한 백엔드 안에서는 번호가 단조증가하니 index를 붙여도 기존 값이 흔들리지 않고, 위 표 3행이 그걸 보여줍니다.

`calls.update(...)`가 `carried`가 아니라 그 회차의 `moved`를 쓰는 것도 맞습니다. `carried`를 썼다면 매 회차마다 누적분 전체가 현재 index로 다시 들어가 반대 방향으로 부풀었을 겁니다.

회귀 테스트도 제가 요청한 모양입니다 — 두 백엔드를 **같은** `0000`으로 주고 2를 못 박습니다. 기존 테스트가 `0000`/`0001`로 갈라줘서 충돌을 못 본 자리를 정확히 메웁니다.

### CI

이 head에서 `freeze-check` pass, `advance-check` pass(2m16s), `pytest`는 작성 시점에 아직 실행 중입니다. 어느 것도 재실행하지 않았습니다. pytest는 위 실행 검증과 **독립**이며 — 저는 테스트가 아니라 함수 자체를 돌렸습니다 — 새 테스트의 기대값(2 calls / 4 files, leaf 2개)은 제 실행 결과와 일치합니다. 머지 전 pytest green 확인만 남습니다.

### 비차단 메모 1건 (수정 요구 아님)

새 docstring(L540)은 근거를 "``backends_built`` holds one backend per task"로 적었는데, 팩토리 쪽 주석(L1480)은 "One object per attempt"입니다. 한 태스크가 두 번 시도되면 리스트에는 백엔드가 2개 들어갑니다. **판정에는 영향이 없습니다** — 그 경우 두 시도의 호출은 `exec_run_calls`에서도 각각 세어지므로 분모가 함께 움직이고, 키가 태스크가 아니라 백엔드 경계에 걸려 있어서 자동으로 맞습니다. 다만 docstring이 내세운 근거가 코드의 실제 근거보다 좁습니다. 다음에 그 문단을 건드릴 일이 생기면 "per attempt"로 맞춰 두면 읽는 사람이 재시도 경로에서 다시 의심하지 않아도 됩니다.

CHANGELOG 항목은 제가 재현한 숫자(6 / 78 / 26)와 문자 그대로 일치하고, 기존 테스트가 왜 못 잡았는지까지 적혀 있습니다. 고칠 것 없습니다.

PR #608, opened 2026-09-17
fix(v2): verify transcript bytes before carrying evidence

hyeonsangjeon:
## Scope

The backend now records a SHA-256 for each generated transcript leaf in its
host-owned output record. Before carrying evidence out of the workspace, it
hashes the byte buffer returned by the descriptor-validated reader and compares
it with the recorded digest. Overwritten output is excluded and its path is
reported with `transcript_sha256_mismatch`.

Normal output bytes, workspace deletion, and repeated `close()` behavior are
preserved. One regression overwrites stdout with same-length bytes and restores
the source after the read, so checking a reopened source path cannot substitute
for checking the copied bytes. It also pins hashes captured at output creation
and preservation of the other transcript files.

The four changed files are the backend, its single regression test,
`CHANGELOG.md` under Unreleased / Fixed, and `tasks/LATEST_TASK_RESULT/README.md`.

## Existing Verification

The single offline target is
`tests/test_transcript_bytes_match_the_host_record.py::test_overwritten_transcript_bytes_are_excluded_but_clean_records_survive`.

- Before the fix: `1 failed`, because overwritten stdout remained in evidence.
- After the fix: `1 passed in 0.36s`.
- `git diff --check` passed during that implementation pass.

These are the existing results. Neither the test nor that diff check was rerun
for this handoff. The backend and test remain byte-identical to their verified
versions. No full suite, real guest run, model call, Azure lookup, or GitHub
project update was performed for this change.

## Reviewed Dependency

#606 at reviewed HEAD `60045f8366997eeee2c47a2793f982ab68071702`.

The English completion record and changelog entry were copyedited with
`/im-not-ai-en`, preserving the recorded evidence and its limits.


hyeonsangjeon (commented):
**APPROVE-CODE**, `806761f0123b96f250236e8870dbe365184d4c47` only. This covers the code at this head. It is not an approval to merge; see the last section.

Reviewed the whole four-file diff in one pass. Reused the three checks already green at this head (`pytest` 06:49:22Z, `advance-check` 06:28:25Z, `freeze-check` 06:26:21Z, all with `head_sha=806761f0123b`) and the recorded regression result, `1 failed` before the fix and `1 passed in 0.36s` after. Nothing was rerun.

### What I checked

**The recorded digest is out of the guest's reach.** `_keep_the_output` hashes `content` from the `payloads` dict built in-process at `batch-runner/core/agentic_v2_microvm_backend.py:871-890`, not a read-back of the file it just wrote. The digest lands in `self.boots`, initialised empty at `:381` and only ever appended to, at `:690`, `:708`, `:738`, `:760`, `:814` and `:840`. Nothing deserializes it. The expected hash is host state, which is the property the fix rests on.

**The check covers the bytes that are actually carried.** `:968` reads into `content`, `:975` hashes that same object, `:989` writes that same object. The path is not reopened between them. `batch-runner/tests/test_transcript_bytes_match_the_host_record.py:34-40` pins exactly this: `read_then_restore` puts the original bytes back at `:37`, immediately after `_read_bytes` returns, so an implementation that hashed a second read of the path would see a match and carry the tampered buffer. That is what makes the recorded `1 failed` load-bearing rather than incidental.

**Ordering holds in both directions.** `:897` records the digest before `:898` appends to `kept`, so every leaf in `kept` has one, and the write-failure branch at `:894-896` breaks before both, so neither is half-populated. On the carriage side the `continue` at `:979` runs before `:989`, so a mismatch leaves no partial file in the evidence directory.

**Mismatch exclusion and clean-output preservation.** The excluded leaf stays out of both the evidence directory and `exec_records_carried`, with `{relative}: transcript_sha256_mismatch` recorded on `exec_records_not_carried`. The test asserts the other five files are carried and byte-identical to their originals.

**Workspace deletion and repeated close are unchanged.** The diff does not touch `close()`. Carriage is guarded by `if not self.closed` at `:944`, and the parent guards the purge with the same flag at `batch-runner/core/agentic_v2_fixture_backend.py:333`, setting it unconditionally at `:349`. A second `close()` enters neither, so there is no duplicate carriage and no second purge. The test compares `exec_records_carried` as a list at `:67-69`, so a duplicated entry would fail it rather than pass under a set comparison.

**A missing digest fails closed.** `(kept.get("sha256") or {}).get(leaf)` yields `None` for a record without one, and `hexdigest() != None` excludes it. Unreachable while `boots` stays in-process, and the right default regardless.

**Security lens finds nothing new.** stdlib `hashlib`, no algorithm selected from input, no secret compared, so comparison timing does not apply, and the traversal guard at `:986` is untouched. `canonical_sha256` is correctly not reused here: it `json.dumps` its argument at `batch-runner/core/agentic_v2_provenance.py:103-112` and would raise on bytes.

**The new `sha256` key breaks no consumer.** No exact-dict assertion on `output_files` exists. `batch-runner/tests/test_agentic_v2_microvm_backend.py:885-886` reads only `["kept"]` and `["not_kept"]`, and `batch-runner/tests/test_the_artifact_keeps_what_the_guest_wrote.py:206` calls `_keep_the_output` itself. Green `pytest` at this head confirms it.

### Non-blocking

1. `exec_records_not_carried` is a single last-write-wins string. `:970` (pre-existing) and `:976` (new) both assign, so if several leaves are tampered with only the last path is named. This is #606 behaviour, the shortfall still shows up in `exec_records_carried`, and the new branch matches the established pattern. Recorded, not asked for.
2. `CHANGELOG.md:343-344` and the completion record cite `60045f8366997eeee2c47a2793f982ab68071702` as the reviewed #606 dependency. That is accurate and still resolves through `refs/pull/606/head`, but #606 landed as squash `6a9e26fa` and its branch is deleted, so `compare` reports `diverged` and a plain clone cannot resolve the SHA. Naming the squash commit beside it would make the citation reachable from history.
3. `tasks/LATEST_TASK_RESULT/README.md` wraps at roughly 80 columns except the `experiment-design` sentence, which runs long.

### Merge is still blocked

#607 landed as main `758eb0e9941ec6483668b5542649d5745a32456d`, and #608 reports `CONFLICTING` / `DIRTY`. Before this can go in:

- Merge current `main` into the branch with a **merge commit**. No rebase, reset, stash or `checkout --`.
- Preserve both #607 `CHANGELOG.md` entries under `### Added`. The two regions are disjoint, #607 at the top of `### Added` and #608 at the top of `### Fixed`, so that file should reconcile without losing either.
- `tasks/LATEST_TASK_RESULT/README.md` is the real reconciliation, and resolving it toward this branch would undo a correction. Main now carries `## Preserved Prior Result: Advance Check Attempt, and Pinning the Azure Resource (2026-08-25)` followed by an "Overtaken in part, 2026-09-17" note recording that the three experiment files show a 2026-09-01 run of five tasks each, finishing 3 of 5, 4 of 5 and 0 of 5, and that the `$0.00` below it is no longer the series total. This branch demotes the same section without the date suffix and without that note, and keeps its "did not run and cost nothing" status line intact. Keep main's dated heading and its note, and place the new Current Task above them.
- Post the reconciliation delta for a final review. It should be small: the merge commit plus the `LATEST_TASK_RESULT` resolution.

hyeonsangjeon (commented):
**FINAL-APPROVE**, `db3e49736f2600ebc2a89086958016f8161601d2`.

Inspected only `806761f0123b..db3e49736f26` and the two parents. The code and test were not reread and nothing was rerun; my APPROVE-CODE for parent `806761f0123b` carries forward, and the blob check below is what lets it.

### The merge itself

`db3e49736f26` has exactly two parents, `806761f0123b96f250236e8870dbe365184d4c47` and `758eb0e9941ec6483668b5542649d5745a32456d`, message `Merge main into b/transcript-byte-integrity-20260917`, tree `508138568784156e9fb0f4daeea7a31764ac8861`, author and committer `hyeonsangjeon <wingnut0310@gmail.com>` at 2026-09-17T09:40:48Z. Two parents means an ordinary merge rather than a rebase or a squash, and the second parent is main at the SHA I named. `compare/806761f0123b...db3e49736f26` reports `ahead_by 2`, `behind_by 0`, so main is now an ancestor.

### Code and test blobs are unchanged

`batch-runner/core/agentic_v2_microvm_backend.py` is blob `db3b24cf36fb` at both `806761f0123b` and `db3e49736f26`. `batch-runner/tests/test_transcript_bytes_match_the_host_record.py` is blob `2543dea2e592` at both. Byte-identical, so the merge introduced no code or test change to review. Against main the PR is still +14 -1 and +71 -0, the same shape I approved.

### CHANGELOG

`+50 -0` against the branch parent and `+10 -0` against main. Zero deletions in both directions, so neither parent's entries could have been dropped. Confirmed by position as well: #607's two `### Added` bullets are at merged lines 15 and 56, under `### Added` at 14, and the transcript entry is the first bullet under `### Fixed` at 385, occupying 386-394 immediately above the #606 entry at 396. `## [Unreleased]` is at 12 with no version heading between 12 and 385, so the entry is still in the unreleased block.

### LATEST_TASK_RESULT

`## Current Task: Transcript Byte Integrity` at 5. The run-place result is summarized as `## Preserved Prior Result: Run-Place Specification (2026-09-17)` at 77, which is the right shape: a 45-line summary that keeps 3 of 5, 4 of 5, 0 of 5, the seven of fifteen denominator, the HTTP 403 refusal read as inability to run rather than a performance result, the six `UNCONTROLLED_DIFFERENCES`, the unmeasured repeat variability, and the `[needs verification]` on the excluded-places count.

The hazard I flagged did not materialize. Merged line 124 carries main's dated heading `## Preserved Prior Result: Advance Check Attempt, and Pinning the Azure Resource (2026-08-25)` and 126-131 carry its `Overtaken in part, 2026-09-17` note verbatim, including the 2026-09-01 run and the statement that the `$0.00` below is no longer the series total. Lines 124 to 752 are byte-identical to main's 107 to 735, all 629 of them, so no preserved section took a collateral edit. `### Result: the run did not happen, and nothing was spent` at 138 and `**Amount spent: $0.00.**` at 140 both survive under that correction rather than being republished uncorrected or deleted.

The 15 deletions against the branch parent are four reflowed lines in Scope and Result, one sentence replaced by three more precise ones, three reworded Skills lines, and the branch's undated heading and `- Status:` bullets giving way to main's dated heading and blockquote. Nothing was weakened. The rerun disclaimer got stricter, from "Neither the test nor that diff check was repeated" to naming the reconciliation's validation as `git diff --check` plus one inspection of the four-file diff.

No PR state and no future merge SHA in either record. The only 40-hex value in the changed region is `60045f8366997eeee2c47a2793f982ab68071702` at line 30, the reviewed #606 dependency, which is a past head rather than a prediction.

### Carried forward, still non-blocking

The `60045f83` citation now appears in both records. It resolves through `refs/pull/606/head` but not in a plain clone, since #606 landed as squash `6a9e26fa` and its branch is deleted. Naming the squash commit beside it would close that. The long `experiment-design` line is now at merged line 64.

This is posted as a comment review rather than a GitHub approval because main has no branch protection, so an approval state would persist onto any later head. The verdict above applies to `db3e49736f2600ebc2a89086958016f8161601d2` and to no other commit.

PR #609, opened 2026-09-17
docs(readme): say where each per-task cost figure is recorded

hyeonsangjeon:
Project #5 card `문제별 문제 풀이 비용과 채점 비용 기록`. Documentation only.

## What this adds

One section in each public README, between the dashboard and development
sections: `Per-task cost records` in `README.md`, `태스크별 비용 기록` in
`README_KR.md`.

It answers two questions that the public READMEs did not answer.

**Where each per-task cost figure lives.** `problem_solving_cost` is recorded on
the inference and report artifacts, written by the run that produced the
deliverable. `grading_cost` is recorded on the grade artifacts, written by the
run that graded it. Neither is part of the other, and this repository publishes
no combined figure.

**What the dashboard actually reads.** Aggregation takes the grade files
directly in `data/grades/` and nothing below that directory. The five
subdirectory names are listed with the reason each is skipped, and the reasons
are not the same:

| Directory | What it holds | Why aggregation skips it |
|---|---|---|
| `_diagnostic/` | Grade records from a run narrowed to a subset, or re-run at a moved fingerprint | Real grade evidence, kept out of the published grade data by policy |
| `_shards/` | Per-shard grade records from a sharded run | Real grades whose tasks can also appear in a file that is already read |
| `_repeats/` | Grade records from deliberate repeats of a run | Real grades that cover the same tasks again by design |
| `_validation/` | Comparison and decision records behind grading choices | Not grade data |
| `_progress/` | Resume state written beside a grade file while it runs | Not grade data |

Only the last two fall outside the grade format. A `_diagnostic/` record is a
real grade and separate evidence, excluded by what this repository publishes
rather than by what the file contains. `_shards/` and `_repeats/` are real
grades for tasks that can be counted elsewhere as well, so combining them with
the files in `data/grades/` can count a task more than once. Nothing here claims
that a sharded run's records have already been merged into that run's file.

Both sections close with a relative link to
`tasks/0828_friday/TASK_PER_TASK_COST_RECEIPTS.md`, which carries the
`cost-receipt-v1` contract.

## What this does not do

**It documents existing behaviour and changes none of it.** No aggregation code,
schema, workflow, test, or data file is touched. A `_diagnostic` grade is still
not aggregated, and the section does not claim otherwise.

No directory count, task count, or cost figure is quoted, so the section does
not go stale as the corpus grows.

## Review and correction

Reviewed content head `409c47418e2bd1f87a95c112b4f44a5a88a9bf60`, review
[...]

PR #610, opened 2026-09-17
fix(hf): bind problem-solving cost receipts to source

hyeonsangjeon:
## Scope

Bind the existing canonical optional `problem_solving_cost` receipt to HF self-report validation. Publication identity retains the receipt, and local validation compares its presence and JSON value before any HF API call.

The diff against main `23c09fe9ce821e00fd49461aa9aeb756fe5b3aae` contains exactly:
- `batch-runner/core/hf_publication.py`
- `batch-runner/tests/test_hf_publication.py`
- `CHANGELOG.md`
- `tasks/LATEST_TASK_RESULT/README.md`

`README.md` and `README_KR.md` match that main baseline byte-for-byte. The completion record preserves the prior README contract, its validation evidence, and reviewed head `aaf554684b8c2e2a95d27f2941e8613b59b48964`.

## Existing validation evidence

- Immutable implementation head: `541c4ea843a1615eb8395860c80c6276ac27652f`
- Integrated head: `ab6a001912898250a6c33d6ed7555cc50ba04c95`
- Targeted selector: `batch-runner/tests/test_hf_publication.py::test_publication_binds_optional_problem_solving_cost_to_source`
- Recorded implementation result: `8 passed in 2.53s`, one executed run. No pytest or CI rerun was requested for record integration.
- `git diff --check 23c09fe9ce821e00fd49461aa9aeb756fe5b3aae` passed. One scope inspection confirmed the four-file diff, unchanged core/test bytes, and identical README bytes.
- Changed, removed, or injected receipts fail with `api.calls == []`. Matching receipts and genuine legacy absence pass. Missing or null source receipts still project to an absent report key; report-side null is not treated as absence.

The required extreme-reasoner was attempted once during implementation but failed because the requested model was unsupported (`model_not_supported`). Compatibility was assessed locally against the existing projector. Completion prose uses the previously loaded `/im-not-ai-en` guidance; experiment-design/report-en/report-ko and UI/animation skills are not applicable.

## Remaining work

- Independent code review and A's final review of the integration delta
- Required CI evidence for the integrated changes

No real HF upload, paid model call, Azure operation, workflow dispatch, full suite, or Project field update was performed.


hyeonsangjeon (commented):
**APPROVE-CODE** at `541c4ea843a1615eb8395860c80c6276ac27652f` (base `f78fe743c4f6f8b0bacc4598115c9c4bd40e000e`, 2 files, +115/−3).

Code only. This PR changes `batch-runner/core/hf_publication.py` and `batch-runner/tests/test_hf_publication.py` and nothing else, so the repository's completion records (`CHANGELOG.md` `[Unreleased]`, `tasks/LATEST_TASK_RESULT/README.md`) are not in it. Those two files sit in #609, which is itself still awaiting a final delta review, so the completion records for this change remain pending behind it. This PR is also still a draft. Nothing below says it is ready to merge or ready for review.

## What the change binds

`problem_solving_cost` reaches `PublicationTaskResult` from the same projection every other bound field comes from. `hf_publication.py:534` computes `projected = project_result_row(...)`, and `hf_publication.py:585` reads `projected.get("problem_solving_cost")` off that. It is never read from the report. A hand-edited `self_report.json` therefore has nothing to bind to except the inference result, which is the property the PR claims.

Optionality survives on both sides. `core/result_projection.py:56-62` omits the key when `project_cost_receipt` returns `None`, and `hf_publication.py:895-897` omits it from the comparison projection on the same condition, so a run predating cost instrumentation still produces no key on either side and gains no null that would read as a recorded zero.

Validation runs before any HF call at source level, not only by the test's assertion. `publish_dataset` validates at `hf_publication.py:1350`, one line before the client exists at `:1353` and four before the first `repo_info` at `:1354`. `publish_dataset_with_receipt` validates at `:1502`, before `:1503` and `:1505`. Both reach the new check through `_validate_publication_files:1122` → `_validate_self_report` → `_validate_self_report_payload:1086`. `assert api.calls == []` is backed by the call order.

## The eight cases

Each one lands where the test says, traced against the source:

| case | decided by |
|---|---|
| `removed-receipt`, `injected-receipt`, `injected-null` | presence check, `hf_publication.py:1021-1027` |
| `changed-receipt`, `receipt-replaced-with-null` | presence matches, `_json_values_equal` fails, generic branch at `:1037-1039` |
| `legacy-absence`, `source-null-projects-absent` | key absent from `projected`, so the loop never sees it |
| `unchanged-receipt` | both sides normalized by `project_cost_receipt`, so the values compare equal |

That covers unchanged, changed, removed, injected, null and absent, with rejection asserted before any API call. I reused your `8 passed in 2.53s` and the clean git-diff-check and reran nothing.

## Compatibility and safety

An ordinary run carrying a receipt is not rejected. `step6_report.py:577` (`entry.update(_task_cost_receipts(r))`) puts the key into `task_results` on the report side, and both sides pass through `project_cost_receipt`, so presence and value agree. That was the one regression worth looking for and it is not present.

`PublicationTaskResult` is `frozen=True`, so the attribute cannot be rebound. The dict is aliased from the loaded inference JSON and is never mutated, and `_task_report_projection` output is only compared, never serialized into an uploaded file.

`dict | None` is not enforced at runtime (`from __future__ import annotations` at `:3`, and dataclasses do not validate annotations). A non-dict JSON value is still value-compared against source, so the binding property holds regardless; `projected.get(...)` types as `Any`, so mypy is satisfied.

## Non-blocking

1. `hf_publication.py:1023-1026` renders exactly the text the generic branch at `:1037-1039` produces for this key, so `pytest.raises(match=...)` cannot tell which of the two checks fired. Four of the five rejecting cases assert the same string through two different code paths. Distinct wording would help if a failure ever has to be triaged from the message alone.
2. The rule is key-specific. The loop at `:1028` iterates `projected.items()` only, so a report row carrying a key the projection does not carry stays unchecked. `grading_cost`, which `_task_cost_receipts` can also emit, is the concrete instance. Pre-existing, not widened here, and outside this PR's stated scope.
3. For the accepted cases the test assigns `task_results[0]["problem_solving_cost"]` directly rather than letting `_build_task_results` carry it, so nothing in this diff proves end to end that the production report writer preserves the value. `step6_report.py:577` does by inspection; no test here binds it.

None of the three changes the verdict.

hyeonsangjeon (commented):
**FINAL-APPROVE** at `ab6a001912898250a6c33d6ed7555cc50ba04c95`.

Scope: the reconciliation delta `541c4ea843a1615eb8395860c80c6276ac27652f..ab6a001912898250a6c33d6ed7555cc50ba04c95` and immutable Git metadata. My APPROVE-CODE for `541c4ea843a1615eb8395860c80c6276ac27652f` ([review 5238055924](https://github.com/hyeonsangjeon/gdpval-realworks/pull/610#pullrequestreview-5238055924)) is reused rather than redone. I reread no unchanged code, reran no test, and dispatched no CI.

## The merge

`ab6a0019` is an ordinary two-parent merge. Parent 1 is `541c4ea843a1615eb8395860c80c6276ac27652f`, the approved code head; parent 2 is `23c09fe9ce821e00fd49461aa9aeb756fe5b3aae`, current `main`. The approved commit is therefore an ancestor, which is what a squash would have cost. Author and committer are `hyeonsangjeon <wingnut0310@gmail.com>` and the message is one line with no trailer.

The compare reports `ahead_by 2, behind_by 0`: `23c09fe9` from `main` and the merge itself. No `batch-runner/` path appears in it.

## The approved code is byte-identical

| path | `541c4ea` / `main` | `ab6a0019` |
|---|---|---|
| `batch-runner/core/hf_publication.py` | `d7a10bba41b0cb719f42b3ccdad15f8b71cd0677` | same |
| `batch-runner/tests/test_hf_publication.py` | `7623df028c11520b1f2e09a2ee2a9beee69a5eec` | same |
| `README.md` | `3610ce149e36dd2d5200d0260e0df546dcea3e74` | same |
| `README_KR.md` | `45f1ac6bb39f7f5b757fafff5d62aa6b8dc5778f` | same |

Blob SHAs read at each ref. The first two match the approved head, so the merge carried no code or test edit. The last two match `main`, so #609's prose survived the merge unchanged in both languages.

The PR diff against `main` is exactly the four expected files: `CHANGELOG.md` (+5/−1), `batch-runner/core/hf_publication.py` (+16/−3), `batch-runner/tests/test_hf_publication.py` (+99/−0), `tasks/LATEST_TASK_RESULT/README.md` (+82/−71).

## Completion records

`CHANGELOG.md:426` is `### Fixed` under `## [Unreleased]` at line 12. Every heading between the two is `###`-level, so no dated release section intervenes. The entry states the defect, the new requirement (matching presence and value before any HF API call), and the two preserved behaviours. It quotes no SHA, date, or merge outcome.

`tasks/LATEST_TASK_RESULT/README.md` opens with `## Current Task: HF Problem-Solving Cost Source Binding`, naming implementation head `541c4ea843a1615eb8395860c80c6276ac27652f`. Scope, outcome, and evidence are present, including the selector that produced `8 passed in 2.53s` and an explicit statement that no rerun was requested during integration and no real HF upload, paid call, Azure operation, or dispatch occurred. Remaining work names the pending final review of this delta and the CI that sits outside the local result. #609 is demoted, not overwritten, to `## Preserved Prior Result` with its reviewed head `aaf554684b8c2e2a95d27f2941e8613b59b48964` — accurate, since [review 5237877289](https://github.com/hyeonsangjeon/gdpval-realworks/pull/609#pullrequestreview-5237877289) was posted against that exact commit, and that head is what merged as `23c09fe9`.

Three claims in the record are checkable from metadata and hold: the four-file scope against `main`, the core and test remaining at the implementation head, and the READMEs retaining `main`'s bytes. Neither record claims this PR's future state, merge SHA, or merge time.

## Non-blocking

1. `tasks/LATEST_TASK_RESULT/README.md` says code review remains pending. Review 5238055924 was submitted at `2026-09-17T15:35:22Z`, nine minutes before this merge commit's `15:44:47Z`, so the line is stale rather than premature. It understates in the safe direction and claims no approval it does not have.
2. The `−1` in the `CHANGELOG.md` diff is a trailing blank line removed at end of file, not part of the entry. Deterministic and harmless.
3. The new entry opens in plain prose where its neighbours under `### Fixed` open with a bold lead sentence. Cosmetic.
4. The three code observations from my earlier review stand unchanged, since the blobs are identical: the presence check at `hf_publication.py:1023-1026` renders the same message as the generic branch at `:1037-1039`; the loop at `:1028` iterates `projected.items()` only, so `grading_cost` stays unchecked; and the accepted-case test assigns `task_results[0]["problem_solving_cost"]` directly rather than exercising `_build_task_results`.

None of the four changes the verdict.

GitHub reports this head MERGEABLE/CLEAN with pytest, advance-check, and freeze-check all SUCCESS; I reused those and started nothing. This is a COMMENTED review, not an approval event: the same account owns the PR, so `event=APPROVE` returns `422 Review Can not approve your own pull request`. The PR is still a draft, and nothing here undrafts or merges it.

PR #635, opened 2026-09-20
feat: bind exact local inputs for the Foundry pilot

hyeonsangjeon:
## Summary

Add an offline input-bundle compiler, materializer and verifier for the fixed
Foundry GPT-5.6 Sol pilot. It reuses the verified config/identity bundles, real
source/reference validators and existing held-parent/no-clobber primitives.

- Preserve the original pinned parquet bytes and publish only the exact
  five-task reference closure. The active cohort requires two reference files.
- Bind task order, prompt/rubric/projection hashes, reference roles/sizes/hashes,
  dataset revision, source pins and upstream ready/reservation identities.
- Reserve an absent destination, publish exclusive directories/files, recheck
  current source and installed bytes, then write ready last. Links, traversal,
  overlap, extras, collisions, drift and partial reuse fail closed without cleanup.
- Split the combined live-input requirement into
  `prepared_input_bytes_unverified` and
  `live_inference_identity_and_wire_unverified`. Explicit verified consumption
  clears only the first; both remain without an input bundle.

The active contract pins 43 sources. Changes to older test files affect only exact
source-count and split-blocker expectations; their selectors were not rerun.

## Validation

The sole local selector collected 135 cases and reported
**1 failed, 134 passed in 71.41s (0:01:11), exit 1**:

```bash
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=batch-runner /usr/bin/python3 -m pytest -q -p no:cacheprovider --tb=short batch-runner/tests/test_gpt56_pilot_input_bundle.py
```

`test_upstream_current_bytes_are_reverified[source-plan]` incorrectly expected an
appended YAML newline to change a canonical JSON contract. The retained case now
changes `pilot.run_id` to `stale-run` and keeps its refusal/no-write assertions.
All 135 cases remain. No production code changed after the invocation, and the
test-only correction has not been rerun. **There is no post-correction local green
result; automatic CI must validate it.**

The cases exercise exact bytes/closure/order, real upstream verification,
no-bundle compatibility, the closed one-blocker transition, stale/tampered data,
links and paths, collisions, publication failures, retained partial state,
mid-publication drift, source/seed isolation and privacy-safe refusals. Offline
guards prohibit subprocess, network, provider auth and model/client/grader
construction. No #631-#634 selector, broad suite or manual workflow was run.

`git diff --check` and `git diff --cached --check` passed. English records and this
PR wording preserve the exact measured result, unexecuted correction and evidence
limits under `im-not-ai-en`.

[...]

PR #634, opened 2026-09-20
feat: materialize inert Foundry pilot config bundles

hyeonsangjeon:
## Scope

Materialize the verified Foundry GPT-5.6 Sol Codex pilot identity into four canonical files: an inert runtime template, prepared-task manifest, grading config and identity linkage. A sibling reservation precedes the files; `pilot-config-bundle-ready.json` is published last. The real identity verifier checks the current contract, source closure and optional evidence before and after publication. Collisions, links, drift and partial reuse fail closed without cleanup.

The fixed five-task order, prompt/reference hashes, Sol/Max/1M requests, timeout/retry/native-cap declarations, one attempt/repeat, fresh sessions and record-only receipt contract are preserved. Grading generation, exp035 configs and production runtime/grader defaults are unchanged.

The grading scope is an inert bundle contract. The historical 220-task `rerun_identity` is omitted because step8 requires a real inference revision there. Future inference repo/revision and the eventual materialized grader source hash remain null. The actual step8 validator checks the emitted grading config. That config alone does not enforce execution-time task selection; no ignored scope keys or fake revision are introduced.

## Leader BLOCK and correction

The leader blocked HEAD `b872446d6e61cfaa8fa1558ee94ea4a6b4ff67ea`: `_runtime()` copied the requested `identity.model` label `gpt-5.6-sol` into `model.deployment`, although the active contract's `foundry_identity.deployment` was null. The artifact passed `ExperimentConfig` validation and could be handed directly to the runner. Outer false launch flags did not enforce the claimed inert boundary. The earlier local result and review did not catch this defect.

The corrected `foundry-pilot-runtime-template-v1` is a closed document with `deployment: null`, `execution: null` and `runnable: false`. Requested model identity remains in the exact sealed dispatch recipe. A small internal shape/parity check replaces runtime-parser acceptance. Explicit null execution prevents the existing parser from supplying runnable defaults for missing blocks. Real `ExperimentConfig.from_dict`, `ExperimentConfig.from_yaml` and the Step 1 CLI entrypoint reject the artifact before data loading, provider settings, route/auth setup or command construction.

Deployment remains null with or without evidence linkage. No model label, placeholder or hash substitutes for an externally reviewed deployment name. A separate future materialization must bind that name into an executable runtime config. `actual_pilot_deployment_not_prepared` remains blocked. The materializer pin was refreshed, so bundles sealed to the older canonical plan are stale. No production runner, grader or workflow file changed in this correction.

## Validation and immutable review

[...]

PR #633, opened 2026-09-20
feat(pilot): seal Foundry dispatch and grading identity

hyeonsangjeon:
## Scope

Add an inert canonical dispatch and grading identity plan for the fixed
Foundry GPT-5.6 Sol Codex five-task pilot, from immutable main
`ed6c64f0afb90b0b3a6a9e4719e44184384296ec`.

- Bind identical ordered dispatch/grading task scopes, prompt/reference hashes,
  dataset revision/parquet hash, source pins, Sol-not-Fast, Max, 1M and existing
  attempt/repeat/timeout/retry/cap declarations.
- Reuse the real step8 validator, source-hash closure, prompt/version and schema
  contracts. Keep the historical 220-task template unchanged and the future
  inference repository/revision null. Emit no argv or runnable config.
- Reverify optional evidence with the real intake verifier and retain only its
  ready SHA256, reviewed source SHA and sealed evaluation time. Publish a
  reservation, canonical plan and ready marker last through existing no-clobber
  and held-parent primitives. Reject drift, links, collisions and partial reuse.
- Let explicit preflight consumption clear only
  `pilot_dispatch_and_grading_identity_not_wired`. Preserve absent/null paths,
  all other required blockers and both false launch flags.

## Validation and immutable review

Exactly one targeted selector ran, with Python `3.10.12` and pytest `9.1.1`:

```bash
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=batch-runner /usr/bin/python3 -m pytest -q -p no:cacheprovider batch-runner/tests/test_gpt56_pilot_identity_plan.py
```

Result: `115 passed in 22.87s`, exit 0. The fixtures prohibit subprocess,
network, provider auth, model/client/grader construction and the grader
entrypoint. They share immutable byte seeds, give each case fresh single-link
files, and call the actual validators without caching their verdicts.
The #631/#632 selectors and broad suites were not rerun.

`first-reviewer` approved implementation HEAD
`0e472bd021bdc868819fc51ac3e0edc5e3a17a4c` with no BLOCK, MAJOR or MINOR
findings and no second-review escalation. It also confirmed the compiler and
preflight pins against Git-object bytes. No code/test correction followed the
passing selector. Final HEAD `176f481f123c951b227a9514eb774cd84a0b9910` adds
only `CHANGELOG.md` and the replacement `tasks/LATEST_TASK_RESULT/README.md`
after that review. Diff whitespace checks passed.

## Evidence boundary and remaining work

[...]

PR #631, opened 2026-09-20
feat: bind Foundry GPT-5.6 Sol pilot evidence offline

hyeonsangjeon:
## Summary

Add a local evidence schema and compile/publish/verify CLI for the active Foundry GPT-5.6 Sol Codex pilot. Six ordered evidence roles bind identity, Max/1M observations, native caps and usage/tariff records to the caller-reviewed source SHA, active plan, exact run and 30 source pins. Requested, documented and observed values remain separate. Unknown or partial evidence cannot receive a ready marker.

Publication reuses the held-parent and atomic no-clobber helpers, checks actual single-link file bytes, retains failed reservations/partial trees and writes `foundry-pilot-evidence-ready.json` last. Closed structured exports exclude raw resource names, endpoints, credentials and personal fields; CLI refusals do not echo input values.

## Validation and review

Base: `d8fd52d9c75687a8e088748c589f9a9a07834a41`.
Immutable implementation: `5a09f59dd4bf8bf5855aa170fa45ec1173fe0fae`.
`first-reviewer`: APPROVE, no BLOCK/MAJOR/MINOR findings. The subsequent commit changes only `CHANGELOG.md` and `tasks/LATEST_TASK_RESULT/README.md`; the implementation remains unchanged.

Exactly one targeted selector ran:

```bash
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=batch-runner /usr/bin/python3 -m pytest -q -p no:cacheprovider batch-runner/tests/test_gpt56_foundry_evidence_intake.py
```

`152 passed in 18.33s`, exit 0. Offline guards forbid subprocess, network, credentials and provider/model/grader construction. No broad suite or selector rerun was used.

## Boundary and remaining work

The bundle establishes local consistency, not Azure authenticity, review approval, served capability, Git lineage or launch authorization. `launch_enabled` and `full_220_enabled` remain false, and every existing preflight blocker remains. No real evidence, Azure/API/CLI, OIDC, provider/model/grader, download, deployment, workflow dispatch or paid execution was used.

Actual evidence acquisition and independent review, native result hosting, deployment and runtime lineage, input/wire evidence, inference publication identity, pilot execution/grading gates and native usage/tariff receipt integration remain outstanding. The fixed five-task contract and historical Copilot record are unchanged. No workflow, `core/qa.py`, HF upload or production runtime/grader defaults changed.

[...]

PR #636, opened 2026-09-20
Bind Foundry deployment identity to the five-task pilot candidate

hyeonsangjeon:
## Scope and outcome

PROJECT5-GPT56-FOUNDRY-DEPLOYMENT-BINDING FULL-9CDA8A92

Bind three private local Foundry account/project/deployment resource-ID files
to complete evidence, using the real #631 evidence, #633 identity, #634 config
and #635 input bundle verifiers. Each original byte digest must match its
evidence subject's `resource_id_sha256`. The candidate derives its deployment
name only from the bound ID's last nonempty path segment.

The new offline CLI/library publishes `pilot-runtime-candidate.json`,
`pilot-deployment-binding.json` and `pilot-deployment-binding-ready.json` with
the existing held-parent and atomic no-clobber primitives. It reserves first,
rechecks upstream and installed bytes, and publishes ready last. Full resource
IDs and private paths stay out of the output and errors. Path/link violations,
secret-like input, stale linkage, digest drift, collisions and partial reuse
fail closed. Failed files and reservations are retained, without automatic
cleanup or retry adoption.

The real runtime parser and validator check the separate candidate. Verified
upstream contracts preserve exact five-task scope, Sol-not-Fast, Max, 1M,
attempt/repeat/session controls, retries/timeouts, native-cap declarations and
record-only receipts. The original inert template stays `deployment: null`,
`execution: null` and `runnable: false`. The active contract now pins 46 sources,
up from 43; the five older test edits update only that count.

Explicit preflight consumption uses a closed mapping to remove only
`actual_pilot_deployment_not_prepared`. No-binding behavior retains the ten
default blockers. Both `launch_allowed` and `full_220_allowed` remain false,
and no launch command is generated.

## Exact validation and immutable review

One targeted invocation, with no rerun:

```bash
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=batch-runner /usr/bin/python3 -m pytest -q -p no:cacheprovider --tb=short batch-runner/tests/test_gpt56_pilot_deployment_binding.py
```

Result: `162 passed in 156.65s (0:02:36)`, exit 0. All upstream verifiers and
the runtime validator are real. Synthetic fixture cases cover raw-ID/privacy,
task/parity, stale/upstream drift, path/link/collision, failed/partial publication,
source immutability, independent copies, the one-blocker transition and false
launch flags. No #631-#635 selector or broad/full suite was rerun.

Base: `9cda8a92040d6679cbfe7f83bca73b0fe657484f`.
Reviewed implementation HEAD: `31813e353d3e98bd666c16102d95bfc3f876be41`.
`first-reviewer` returned APPROVE with no BLOCK, MAJOR or MINOR findings and no
second-review escalation. `llm-systems-engineer` approved the same HEAD without
blocking or high-confidence findings. Both reviews were read-only. The later
CHANGELOG and completion-record commit is outside that implementation boundary.
`git diff --check` and `git diff --cached --check` passed.

## Evidence boundary and remaining work

This is `offline_deployment_resource_binding_not_launch_or_execution`. The
candidate is parser-compatible, not launch-authorized. Its outer false flags do
not enforce runtime admission, and Step 2 does not automatically consume this
binding. Local hashes do not authenticate Azure facts, prove served capability
or attest actual input consumption. Only synthetic IDs/evidence were used.

Live inference identity, input consumption and wire binding; native sandbox and
result hosting; actual candidate deployment and pilot execution; execution-time
gating/native caps; grading run/source identity; and usage/tariff receipts remain
unresolved. No Azure API/CLI calls, credential/OIDC lookup, provider/model/client
execution, inference, grading, download, workflow dispatch or paid execution
occurred. Grading and runtime defaults, workflows, `core/qa.py`, HF upload, exp035 and historical
ledger/evidence bytes are unchanged. Project and merge remain with the leader.

The full skill catalog was checked once. `experiment-design` constrained the
work to local byte binding; `im-not-ai-en` was applied to the English records.
`extreme-reasoner` does not apply because its protected files are unchanged.
UI/animation skills and a new grading-role task are outside this backend scope.


hyeonsangjeon (commented):
FINAL-APPROVE at 3e4846f846451daed91aa387003c9c18a943c3a8

Leader review covered the immutable private Foundry resource-ID parser, exact evidence hash and hierarchy binding, deployment-leaf-only runtime candidate, four upstream verifier chain, one-blocker preflight transition, 162-case selector evidence, immutable llm-systems and first-reviewer boundaries, and completion records. Validate, pytest, comparison-contracts, and advance-check passed at this exact HEAD; no blocking findings remain. Launch flags remain false and no credentials, endpoints, or commands are emitted.

PR #638, opened 2026-09-20
feat: bind pilot results to Codex transport receipts

hyeonsangjeon:
## Summary

Add a session-owned transport receipt and result gate for the registered
Foundry GPT-5.6 Sol five-task pilot. The pinned Codex SDK exposes app-server
stdio, correlated replies and native thread usage. It does not expose Foundry
HTTP requests or served-model identity; those fields remain explicitly
unavailable, and `live_inference_identity_and_wire_unverified` remains present.

- Hash the exact serialized request bytes after a complete write and flush.
- Bind run/task/attempt identity, prepared capture, upstream bundles and the
  requested route; privately compare the endpoint account with the reviewed
  account-resource digest and leaf before workspace/auth/client startup.
- Require current bytes and the session's own observation before result/file
  acceptance. Publish receipts without clobbering and the ready marker last.
- Preserve native usage and missing/null states without prices, raw text,
  resource IDs, endpoints, credentials, private host paths or exception strings.
- Preserve legacy absent/null behavior and all false launch/full-220 flags.

Source pins and grader closure metadata are refreshed for changed core bytes.
Six existing test files update only their exact source-count expectations.
No workflow, grader production, `core/qa.py` or HF upload code changes.

## Validation and immutable review

Exactly one local pytest command ran from `batch-runner`:

```bash
/ai-work/venvs/gdpval-realworks-py310/bin/python -m pytest -q tests/test_gpt56_pilot_wire_receipt.py --tb=short
```

`88 passed in 642.32s (0:10:42)`, exit 0, at initial implementation
`59aa7f00de5505e0ed5c1bff8e8ae7757ff9398a`. Both reviewers identified an account
binding BLOCK. The correction binds the current private resource bytes to the
sealed capture and adds a wrong-account/same-deployment regression before the
runtime boundary. The resulting 89-case selector was not rerun locally;
automatic CI is required. Full Backend timing is unmeasured.

Both `llm-systems-engineer` and `first-reviewer` returned APPROVE on fixed
implementation `89e9a33baa3b6c4b3c404f7c2c6f61798003a868`, against base
`0cd4c5e76805384a77afabf28b3663a3ad6596c0`, with no remaining blocking findings.
These are read-only static reviews, not test or launch approval. Completion
records are a subsequent records-only commit. `git diff --check` passed.

## Evidence limits and remaining work

Synthetic fixtures prove verifier behavior only. No Azure/HF query,
credential/OIDC lookup, provider/model/client/grader execution, inference,
[...]

PR #639, opened 2026-09-20
feat: add session-owned native result host receipts

hyeonsangjeon:
## Scope

Add session-owned native-workspace and Step 2 accepted-result receipts for the registered Foundry GPT-5.6 Sol five-task pilot. This starts from main `f6d76e2084bcda75de5dad15b98503f598734bc0` in a new worktree.

- Attach the host witness to the exact live capture/wire session. Observe workspace containment and the selected app-server policy before cleanup; reject unsafe files, altered handoffs, drift, partial reuse and forged saved results.
- Bind the actual accepted rows and saved deliverable bytes, with a non-circular result fingerprint and atomic no-clobber, ready-last publication. Public host receipts contain relative file identities and digests, not private content or host/resource data.
- Preserve legacy absent/null behavior. Extend the existing wire selector instead of adding a test file. Refresh the directly affected 55-file source closure and completion records; leave workflows, the three-way Backend partition, QA, grader production and HF upload unchanged.

## Validation and immutable review

Exactly one targeted pytest command ran:

```bash
PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=batch-runner /usr/bin/python3 -m pytest -q -p no:cacheprovider --tb=short batch-runner/tests/test_gpt56_pilot_wire_receipt.py -k native_result_host
```

Result: `1 failed, 70 passed, 84 deselected in 2790.26s (0:46:30)`, exit 1. Pytest collected 155 cases and selected 71.

The failed happy-path test reached its preflight assertion after the receipt checks, then called nonexistent `pilot.inspect`. It now calls `pilot.inspect_plan`. Static cost inspection also found that open YAML text streams bypassed the existing fixture parse cache; the cache now rereads current contents and returns fresh parsed copies while retaining real validators and drift checks. Neither correction was rerun. No local green result, measured speedup or hosted CI headroom is claimed. Fresh automatic CI must supply that evidence.

`git diff --check` and staged diff checks passed. Read-only `llm-systems-engineer` and `first-reviewer` reviews both returned APPROVE with no remaining findings on implementation HEAD `d2850e6dd4d12ab5180bdaaeef17c840a2a90fa3`. Final HEAD `3d8336933ed26266332c97b3caf1235b95cb98e4` adds only the completion records after that review boundary.

The records retain PR #638's original `88 passed in 642.32s (0:10:42)` as pre-account-correction evidence and the corrected 89-case full selector as not locally rerun. Historical hosted Backend run `35520193108` passed all three jobs at `9e6e42b85e56478f8082cd7bc9ae1fb6feb31812`; that does not validate this code.

## Evidence limits and remaining work

No live pilot occurred. `native_sandbox_and_result_bundle_host_unverified` and `live_inference_identity_and_wire_unverified` remain unresolved. Only a future runtime-owned verified host bundle may satisfy the native-host blocker; there is no disk-only witness adoption or offline preflight option that converts synthetic files into live evidence. The receipt does not attest remote isolation, served identity, native caps or usage/tariff/billing.

All launch/full-220 flags remain false, and no runnable command was added. Actual evidence acquisition and separately approved deployment/execution, live wire/served identity, real native-host receipts, cap enforcement and usage/tariff receipts remain separate work. No Azure/HF/OIDC/credential lookup, provider/model/client/grader execution, inference, download, paid execution or manual workflow ran. No Project or merge action was taken.


hyeonsangjeon (commented):
FINAL-APPROVE at 2e36d668a64101d379d64f27064d9e675118b3f5

Leader immutable-head review found no blocking issue. The native-host boundary remains session-owned and non-launching, completion records stop at pre-merge facts, and pytest, comparison-contracts, pilot-contracts, native-host-contracts, validate, advance-check, freeze-check, and containment checks are green at this exact HEAD.

